PR #5, opened 2020-09-04
Add Validation

### LINKED ISSUE #4: Ensure valid JSON on websocket message
In [api.ts](https://github.com/MyCryptoHQ/signer/blob/master/src/api/api.ts#L14), we need to do a sanity check to ensure the message is a valid JSON string as we can currently DoS the signer by sending an invalid JSON string to throw an uncaught exception

Screenshot 2020-09-03 at 16 49 58

Once this exception has been thrown, the ws won't take any more incoming messages

Note: I know it's still in development, but making an issue anyway to ensure it's fixed :D

FrederikBolding:
Fixes #4

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -1,10 +1,6 @@
-import { JsonRPCRequest, JsonRPCResponse } from '@types';
+import { JsonRPCRequest, JsonRPCResponse, SUPPORTED_METHODS } from '../types';
 import { safeJSONParse } from '../utils';
-
-const SUPPORTED_METHODS = {
-  SIGN_TRANSACTION: 'eth_signTransaction',
-  ACCOUNTS: 'eth_accounts',
-};
+import { isValidRequest } from './validators';
 
 export const handleRequest = (
   data: string,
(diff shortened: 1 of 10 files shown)


src/api/util.ts
```diff
@@ -8,3 +8,12 @@ export const makeTx = (request: JsonRPCRequest): TransactionRequest => {
 export const addHexPrefix = (str: string) => {
   return str.slice(0, 2) === "0x" ? str : "0x" + str;
```
vilsbole, line 9: interpolation

src/api/validators.ts
```diff
@@ -0,0 +1,63 @@
+import { validator } from "@exodus/schemasafe";
+import { SUPPORTED_METHODS, JsonRPCRequest } from "../types";
```
vilsbole, line 2: aliases

src/api/api.ts
```diff
@@ -1,27 +1,40 @@
...
+    reply({
+      id: null,
+      jsonrpc: "2.0",
+      error: { code: "-32700", message: "Parse error" },
+    });
+    return;
+  }
+  const request = json[1] as JsonRPCRequest;
+  if (Object.values(SUPPORTED_METHODS).includes(request.method)) {
+    if (!isValidRequest(request)) {
```
vilsbole, line 22: No nested conditionals

vilsbole (changes requested):

vilsbole (approved):

PR #7, opened 2020-09-23
Add Tray icon

FrederikBolding:


src/main.ts
```diff
@@ -27,19 +39,98 @@ const createWindow = (): void => {
...
+    };
+  }
+
+  // TASKBAR BOTTOM
+  // Since all the other cases were handled, we can be sure we're dealing with a bottom taskbar
+  return {
+    x: Math.round(trayBounds.x + trayBounds.width / 2 - WIDTH / 2),
+    y: workArea.height - HEIGHT,
+  };
+};
```
GuillaumeRx, line 98: could this be a switch ?

FrederikBolding: Hmm, how would you suggest? @GuillaumeRx

GuillaumeRx:
Capture d’écran 2020-09-28 à 12 24 41
On mac it's a bit weird

src/main.ts
GuillaumeRx: After some testing it can't be a switch, good as it is now

src/main.ts
```diff
@@ -27,19 +39,98 @@ const createWindow = (): void => {
...
+  window.show();
+  window.focus();
+};
+
+const toggleWindow = () => {
+  if (window.isVisible()) {
+    window.hide();
+  } else {
+    showWindow();
+  }
```
GuillaumeRx, line 114: you can use ternary operators here

src/main.ts
```diff
@@ -27,19 +39,98 @@ const createWindow = (): void => {
...
+};
+
+const createTray = () => {
+  tray = new Tray(path.join(__dirname, 'favicon.png'));
```
GuillaumeRx, line 118: `favicon.png` should be a 32x32px so it renders well in macos

GuillaumeRx (changes requested):

src/main.ts
FrederikBolding: I don't understand how you would use ternaries here, can you explain?

FrederikBolding:
I think the new favicon could fix the MacOS issue as well, please check 👍

src/main.ts
GuillaumeRx:
```ts
window.isVisible() ? window.hide() : showWindow()
```

GuillaumeRx:
Capture d’écran 2020-09-28 à 17 16 17
Still a bit big 🤔

FrederikBolding:
Added a 16x16 version for Mac OS @GuillaumeRx

GuillaumeRx (approved):
Good now 👌

PR #14, opened 2020-10-01
Standardize bridge API to improve typing

FrederikBolding:


FrederikBolding:
[ch7052]

src/api/api.ts
```diff
@@ -19,8 +20,7 @@ const requestSigning = (
   return new Promise((resolve, _reject) => {
     webContents.send(IPC_CHANNELS.API, request);
 
-    ipcMain.on(IPC_CHANNELS.API, function _listener(_event, arg) {
-      const response = arg as Omit<JsonRPCResponse, 'jsonrpc'>;
+    ipcBridgeMain(ipcMain).api.on(function _listener(_event, response) {
```
Mrtenz, line 23: Can we use an arrow function here, and remove the `_`?

Mrtenz (changes requested):
Would be cool if we can fully type JSON RPC requests and responses as well, but that may be out of scope for this PR.

src/api/api.ts
FrederikBolding: It has to be a named function as otherwise we can't remove the listener later. But I can remove the `_`

FrederikBolding:
> Would be cool if we can fully type JSON RPC requests and responses as well, but that may be out of scope for this PR.

Agreed!

Mrtenz (approved):

PR #16, opened 2020-10-02
Remove app frame

FrederikBolding:


src/main.ts
```diff
@@ -115,7 +118,11 @@ const toggleWindow = () => (window.isVisible() ? window.hide() : showWindow());
 
 const createTray = () => {
   tray = new Tray(path.join(__dirname, 'favicon.png'));
-  tray.on('right-click', toggleWindow);
+  const contextMenu = Menu.buildFromTemplate([
```
GuillaumeRx, line 121: could this be on right click ? Right now on mac it opens the context menu on left-click also

src/main.ts
```diff
@@ -29,6 +29,9 @@ const createWindow = (): void => {
     height: HEIGHT,
     fullscreenable: false,
     resizable: false,
+    // Spectron fails randomly when the app is frameless
+    // We don't use NODE_ENV as it is controlled by electron-forge
+    frame: process.env.isTest === 'true',
```
GuillaumeRx, line 34: Can we set it to `IS_DEV` env variables should be uppercase

GuillaumeRx (changes requested):

GuillaumeRx (approved):
LGTM !

PR #19, opened 2020-10-07
Add accounts to redux state

FrederikBolding:


FrederikBolding:
[ch7109]

src/api/db.ts
```diff
@@ -12,7 +12,7 @@ const init = (password: string) => {
   try {
     store = new Store({ encryptionKey: password, clearInvalidConfig: true });
     // Write something to the store to actually create the file
-    store.set('accounts', []);
+    store.set('accounts', {});
```
Mrtenz, line 15: Wouldn't this override the accounts every time `init` is called? Maybe check if `accounts` exists before writing?

src/app/store/persistor.ts
```diff
@@ -0,0 +1,9 @@
+import { EnhancedStore } from '@reduxjs/toolkit';
+import { persistStore } from 'redux-persist';
+
+import { ApplicationState } from '.';
+
+export const createPersistor = (store: EnhancedStore<ApplicationState>) =>
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
Mrtenz, line 8: Why?

src/app/store/utils/storage.ts
```diff
@@ -0,0 +1,20 @@
+import { getAccounts, setAccounts } from '@app/services';
+
+export const storage = {
+  getItem: (key: string) => {
+    if (key === 'accounts') {
+      return getAccounts();
+    }
+    throw new Error('Invalid key');
+  },
+  setItem: (key: string, value: any) => {
```
Mrtenz, line 10: Can you replace `any` with `unknown` here?

Mrtenz (changes requested):

src/api/db.ts
FrederikBolding: `init` is only for initializing the store and will delete anything in the store anyway!

src/app/store/persistor.ts
FrederikBolding: For some reason the `redux-persist` typing doesn't include `manualPersist`

Mrtenz (approved):

PR #22, opened 2020-10-08
 Basic Add Account Logic

FrederikBolding:


FrederikBolding:
[ch7100]

src/api/api.ts
```diff
@@ -36,15 +37,18 @@ const requestSigning = (
...
       return requestSigning(request, webContents);
-    // @todo Actual account handling
+    // @todo Dont expose all accounts at once?
+    // @todo Decide whether to fetch directly from store?
     case SUPPORTED_METHODS.ACCOUNTS:
       return toJsonRpcResponse({
         id: request.id,
-        result: ['0x82D69476357A03415E92B5780C89e5E9e972Ce75']
+        result: Object.values(await getAccounts())
+          .filter((a) => a.address)
```
Mrtenz, line 50: Is this necessary? In what situation would `address` be `undefined` or `null`?

src/api/crypto.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { IPC_CHANNELS } from '@config';
+import { CryptoRequestType } from '@types';
+
+import { handleRequest, runService } from './crypto';
+
+jest.mock('@ethersproject/wallet', () => ({
+  Wallet: jest.fn().mockImplementation(() => ({
+    address: '0x4bbeEB066eD09B7AEd07bF39EEe0460DFa261520',
+    signTransaction: jest.fn().mockImplementation(() => Promise.resolve('signed'))
```
Mrtenz, line 11: Can we return an actually signed transaction here? That may prevent situations in the future where some library depends on a proper signed transaction hexadecimal string, but gets "signed" instead, something that broke a bunch of stuff in legacy.

src/app/screens/AddAccount.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { getAddressFromPrivateKey } from '@app/services/WalletService';
+import { ApplicationState, createStore } from '@app/store';
+
+import { AddAccount } from '.';
+
+jest.mock('@app/services/WalletService', () => ({
+  getAddressFromPrivateKey: jest
+    .fn()
+    .mockImplementation(() => Promise.resolve({ uuid: 'uuid', address: 'address' }))
```
Mrtenz, line 16: Return actual UUID/address. See comment above.

src/types/route.ts
```diff
@@ -0,0 +1,5 @@
+export interface Route {
+  path: string;
+  component: React.ComponentType<any>;
```
Mrtenz, line 3: `any` -> `unknown`

Mrtenz (changes requested):

src/api/api.ts
FrederikBolding: It seems that some metadata is stored by `persist-redux`, maybe I should filter it out some other way. Hmm

Mrtenz (approved):

PR #24, opened 2020-10-12
Save Private Keys

FrederikBolding:


FrederikBolding:
[ch6992]

src/api/secrets.ts
```diff
@@ -0,0 +1,47 @@
...
+import keytar from 'keytar';
+
+import { IPC_CHANNELS, KEYTAR_SERVICE } from '@config';
+import { SecretsRequest, SecretsRequestType, SecretsResponse, TUuid } from '@types';
+import { decrypt, encrypt, hashPassword } from '@utils/encryption';
+
+// @todo STORES HASHED PASSWORD FOR ENCRYPTION - THINK ABOUT THIS
+let encryptionKey: string;
+
+const SetEncryptionKey = (key: string) => (encryptionKey = hashPassword(key));
```
Mrtenz, line 11: Function names should start with a lowercase letter (`camelCase`). Here and for the other functions as well.

src/config.ts
```diff
@@ -10,3 +11,5 @@ export const SUPPORTED_METHODS = {
 };
 
 export const WS_PORT = 8000;
+
+export const KEYTAR_SERVICE = 'MYC_Signer';
```
Mrtenz, line 15: Maybe use the full application name here, as this may be shown in GUI applications reading from the OS key chain (like the built-in macOS app).

src/utils/encryption.ts
```diff
@@ -0,0 +1,18 @@
...
+};
+
+export const encrypt = (data: string, key: string) => {
+  const aes = new ModeOfOperation.ctr(utils.hex.toBytes(key));
+  const encryptedBytes = aes.encrypt(utils.utf8.toBytes(data));
+  return utils.hex.fromBytes(encryptedBytes);
+};
+
+export const hashPassword = (password: string) => {
+  return utils.hex.fromBytes(SHA256(password, { asBytes: true }));
```
Mrtenz, line 17: SHA256 hashes are easy to compute, potentially making bruteforcing easier, so it shouldn't be used to derive a cryptographic key. Consider using something like [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2) with a high number of iterations.

Mrtenz (changes requested):

src/utils/encryption.ts
FrederikBolding: We use SHA256 on beta, which is why I initially just used it. I've reached out to Harry for a recommendation on what to use.

Mrtenz (approved):

PR #30, opened 2020-10-20
Wallet Creation + Basic Mnemonic Support

FrederikBolding:
- [x] Mnemonic Wallet Creation
- [x] Add Mnemonic Account
- [x] Sign with Mnemonic Account
- [x] Remaining tests


FrederikBolding:
[ch7187]

jest_config/__fixtures__/mnemonic.ts
```diff
@@ -0,0 +1,2 @@
+export const fMnemonicPhrase =
```
Mrtenz, line 1: Can we use something like `test test test test test test test test test test test test test test test test test test test test test test test blade` (yeah that's a valid one) here just to make sure that nobody will ever use this mnemonic?

src/api/crypto.ts
```diff
@@ -1,39 +1,115 @@
...
+  password?: string;
+  offset?: number;
+  limit: number;
+}) => {
+  const rootNode = HDNode.fromMnemonic(phrase, password);
+  const addresses = [];
+  for (let i = 0; i < limit; i++) {
+    const index = i + offset;
+    const dPath = `${dPathBase}/${index}`;
+    const node = rootNode.derivePath(dPath);
```
Mrtenz, line 76: It may be slightly faster to derive the base master node (`dPathBase`) once, then deriving the child keys from that.

```suggestion
  const baseNode = rootNode.derivePath(basePath);
  for (let i = 0; i < limit; i++) {
    const index = i + offset;
    const node = baseNode.derivePath(`${index}`);
```

src/api/crypto.ts
```diff
@@ -1,39 +1,115 @@
...
-      const { privateKey } = request;
-      return getAddress(privateKey);
+      const { args, wallet } = request;
+      if (wallet === WalletType.PRIVATE_KEY) {
+        return getPrivateKeyAddress(args as string);
+      } else if (wallet === WalletType.MNEMONIC) {
+        if ((args as GetMnemonicAddressArgs).dPath) {
+          return getMnemonicAddress(args as GetMnemonicAddressArgs);
+        }
+        return getMnemonicAddresses(args as GetMnemonicAddressesArgs);
```
Mrtenz, line 102: Can you infer the type of args here based on `wallet`?

src/app/screens/AddAccount/AddAccountMnemonic.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { useState } from 'react';
+
+import { useHistory } from 'react-router-dom';
+
+import { useAccounts } from '@app/hooks';
+import { ROUTE_PATHS } from '@app/routing';
+import { getAddress } from '@app/services';
+import { GetMnemonicAddressesResult, WalletType } from '@types';
+
+const dPathBase = "m/44'/60'/0'/0";
```
Mrtenz, line 10: Is support for other derivation paths planned in the future?

src/app/screens/SignTransaction/SignTransaction.tsx
```diff
@@ -0,0 +1,65 @@
...
+      )}
+      {currentAccount && currentAccount.type === WalletType.MNEMONIC && (
+        <SignTransactionMnemonic
+          hasPersistentPrivateKey={hasPersistentPrivateKey}
+          onDeny={handleDeny}
+          onAccept={handleAccept}
+          tx={formattedTx}
+          currentAccount={currentAccount}
+        />
+      )}
```
Mrtenz, line 60: I think this pattern needs to be thought about, and if possible should be made more abstract. We're doing the same thing in core currently, and it causes issues with scalability etc. Probably out of scope for this PR.

(Here and in some other files as well)

src/app/screens/SignTransaction/SignTransactionMnemonic.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const changeMnemonicPhrase = (e: React.ChangeEvent<HTMLInputElement>) =>
+    setPhrase(e.currentTarget.value);
+
+  const changePassword = (e: React.ChangeEvent<HTMLInputElement>) =>
+    setPassword(e.currentTarget.value);
+
+  const handleAccept = async () => {
+    if (!hasPersistentPrivateKey) {
+      const { privateKey } = (await getAddress({
```
Mrtenz, line 24: This will throw an uncaught error if the mnemonic is invalid. Form input should be checked, and promise rejections should be handled.

Mrtenz (changes requested):

src/app/screens/AddAccount/AddAccountMnemonic.tsx
FrederikBolding: Yeah I would say we should do that.

Mrtenz (approved):

PR #37, opened 2020-11-03
UI Setup

GuillaumeRx:
## Content

* Add styled-components, styled-system and theme-ui
* Create theme, variants and globalStyle
* Add basic CSS rules to theme
* Create default typographies


FrederikBolding (changes requested):
Does not work with production CSP:

`Refused to apply inline style because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'style-src' was not explicitly set, so 'default-src' is used as a fallback.`

FrederikBolding (approved):

PR #49, opened 2021-03-03
Top Navigation

FrederikBolding:


FrederikBolding:
[ch8258]

src/app/components/Core/Box.tsx
```diff
@@ -0,0 +1,48 @@
...
+  color,
+  layout,
+  flexbox,
+  flexVariants,
+  display,
+  border,
+  shadow
+);
+
+export default Box;
```
GuillaumeRx, line 48: I don't think we need this one since rebass has his own `Box` and `Flex` components

src/app/components/Core/LinkApp.tsx
```diff
@@ -0,0 +1,206 @@
...
+    '&:active': {
+      opacity: 1
+    },
+    '&:focus': {
+      opacity: 1
+    }
+  }
+};
+
+const SLink = styled('a')<LinkStyleProps & React.AnchorHTMLAttributes<HTMLAnchorElement>>`
```
GuillaumeRx, line 96: We should convert to the Rebass `Link` and get rid of the style-component/styled-system method

src/app/theme/theme.ts
```diff
@@ -53,10 +55,54 @@ export const theme: DefaultTheme = {
...
+    display: 'flex',
+    flexDirection: 'column',
+    ...FLEX_RECIPES.center
+  }
+};
+
+export type FlexVariants = keyof typeof FLEX_VARIANTS;
+export const flexVariants = variant({
+  variants: FLEX_VARIANTS
+});
```
GuillaumeRx, line 108: Migrate those variant to the theme, with rebass it's easier to ship them with the theme (https://rebassjs.org/variants)

GuillaumeRx (changes requested):

GuillaumeRx (approved):

PR #51, opened 2021-03-04
Home Screen

FrederikBolding:


FrederikBolding:
[ch8263]

src/app/components/Core/TimeElapsed.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useState } from 'react';
+
+import { formatDistanceToNow } from 'date-fns';
+
+import { useInterval } from '@app/hooks';
+
+const TimeElapsed = ({ value }: { value: number }) => {
+  const formatTimeDifference = (val: number) =>
```
Mrtenz, line 8: Maybe move this to a separate test file.

src/app/components/TxHistory/TxHistoryCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+          pr="14px"
+          color="BLUE_GREY"
+          sx={{ textDecoration: isApproved ? 'none' : 'line-through' }}
+        >
+          <TimeElapsed value={item.timestamp} />
+        </Body>
+        <LinkApp
+          href={ROUTE_PATHS.TX}
+          data-testid={`select-tx-history`}
+          onClick={() => dispatch(selectTransaction(item))}
```
Mrtenz, line 33: Don't use arrow functions in props.

src/app/components/TxQueue/TxQueueCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <Image src={waiting} height="20px" width="20px" mr="8px" />
+        <Body color="PURPLE">WAITING ON ACTION</Body>
+      </Box>
+      <Box variant="rowAlign" mt="2">
+        <FromToAccount sender={tx.from as TAddress} recipient={tx.to as TAddress} />
+        <LinkApp
+          href={ROUTE_PATHS.TX}
+          ml="auto"
+          data-testid={`select-tx-${item.id}`}
+          onClick={() => dispatch(selectTransaction(item))}
```
Mrtenz, line 29: Don't use arrow functions in props.

src/app/components/TxQueue/TxQueueCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const { tx } = item;
+
+  return (
+    <Box py="16px">
+      <Box variant="rowAlign">
+        <Image src={waiting} height="20px" width="20px" mr="8px" />
+        <Body color="PURPLE">WAITING ON ACTION</Body>
+      </Box>
+      <Box variant="rowAlign" mt="2">
+        <FromToAccount sender={tx.from as TAddress} recipient={tx.to as TAddress} />
```
Mrtenz, line 24: Can we change the type of `tx.from` and `tx.to` so these type casts aren't necessary?

src/app/hooks/useInterval.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useRef } from 'react';
+
+// From https://overreacted.io/making-setinterval-declarative-with-react-hooks/
+export function useInterval(
+  callback: () => void,
+  delay: number,
+  immediate?: boolean,
+  trigger?: any
```
Mrtenz, line 8: Use strict type here.

src/app/screens/SignTransaction/SignTransaction.tsx
```diff
@@ -33,48 +32,44 @@ export const SignTransaction = () => {
...
-      {currentAccount && currentAccount.type === WalletType.KEYSTORE && (
+      {isWaiting && currentAccount && currentAccount.type === WalletType.KEYSTORE && (
         <SignTransactionKeystore
           onAccept={handleAccept}
           onDeny={handleDeny}
-          tx={formattedTx}
+          tx={tx}
           currentAccount={currentAccount}
         />
       )}
```
Mrtenz, line 75: We can simplify this by moving the components to a dictionary or function, since the props are the same across all components anyway.

src/app/screens/Transaction.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+import { SignTransaction } from './SignTransaction';
+
+export const Transaction = () => (
+  <>
+    <SignTransaction />
+  </>
```
Mrtenz, line 8: No need to wrap this in a fragment.

src/utils/index.ts
```diff
@@ -16,4 +10,5 @@ export const safeJSONParse = <T>(str: string): [null, T] | [Error, null] => {
```
Mrtenz, line 10: Can we move the utils here to a separate file, not in `index.ts`?

src/utils/truncate.ts
```diff
@@ -0,0 +1,3 @@
+export const truncate = (children: string): string => {
+  return [children.substring(0, 6), '…', children.substring(children.length - 4)].join('');
```
Mrtenz, line 2: You can use a template string here rather than an array with `join()`.

Mrtenz (changes requested):

src/app/components/TxQueue/TxQueueCard.tsx
FrederikBolding: I would prefer using the `ethers` type for transactions

src/app/components/Core/TimeElapsed.tsx
FrederikBolding: What do you mean?

Mrtenz: Sorry, I meant separate file, and add some tests to it. That makes it easier to swap `date-fns` out with something else, if we ever need to do so.

src/app/components/TxQueue/TxQueueCard.tsx
Mrtenz: The problem is that we will need to type cast `from` and `to` every time we use it with a function that expects a `TAddress`. We could do declaration merging to change the type in `ethers` itself if we want to keep using the `ethers` type.

src/app/components/TxHistory/TxHistory.tsx
```diff
@@ -10,8 +11,8 @@ export const TxHistory = ({ history }: { history: TxHistoryEntry[] }) => (
     {history
       .slice()
       .sort((a, b) => b.timestamp - a.timestamp)
-      .map((h, i) => (
-        <React.Fragment key={i}>
+      .map((h) => (
+        <React.Fragment key={generateUUID()}>
```
Mrtenz, line 15: I don't think this will work, since it will re-generate the UUID every time the component is rendered. That may have similar or worse performance issues as compared to using the index.

src/app/components/TxQueue/TxQueue.tsx
```diff
@@ -1,14 +1,15 @@
...
 
 import { Divider } from '../Divider';
 import { TxQueueCard } from './TxQueueCard';
 
 export const TxQueue = ({ queue }: { queue: TxQueueEntry[] }) => (
   <>
-    {queue.map((q, i) => (
-      <React.Fragment key={i}>
+    {queue.map((q) => (
+      <React.Fragment key={generateUUID()}>
```
Mrtenz, line 12: See comment above.

src/app/screens/Accounts.tsx
```diff
@@ -14,8 +14,8 @@ export const Accounts = () => {
     <>
       <Link to={ROUTE_PATHS.HOME}>Back</Link>
       <br />
-      {accounts.map((a, index) => (
-        <div key={index}>
+      {accounts.map((a) => (
+        <div key={a.address}>
```
Mrtenz, line 18: Is the address guaranteed to be unique?

src/app/screens/AddAccount/AddAccountMnemonic.tsx
```diff
@@ -77,8 +77,8 @@ export const AddAccountMnemonic = () => {
       {addresses.length > 0 && (
         <>
           <br />
-          {addresses.map((address, index) => (
-            <button key={index} onClick={() => addAddress(address)}>
+          {addresses.map((address) => (
+            <button key={address.address} onClick={() => addAddress(address)}>
```
Mrtenz, line 81: See comment above.

Mrtenz (changes requested):

Mrtenz (approved):

PR #57, opened 2021-03-12
Add Account Flow v2

FrederikBolding:


FrederikBolding:
[ch8259]

src/app/store/account.slice.ts
```diff
@@ -64,37 +77,51 @@ export const getAccounts = createSelector(
...
-      wallet
-    });
+export function* fetchAccountsWorker({
+  payload: wallets
+}: PayloadAction<(SerializedWallet & { persistent: boolean })[]>) {
+  const accounts: IAccount[] = yield select(getAccounts);
+
+  try {
+    for (let i = 0; i < wallets.length; i++) {
+      const wallet = wallets[i];
```
Mrtenz, line 92: This can be simplified a bit
```suggestion
    for (const wallet of wallets) {
```

src/app/store/account.slice.ts
```diff
@@ -64,37 +77,51 @@ export const getAccounts = createSelector(
...
+      const address: TAddress = yield call(ipcBridgeRenderer.crypto.invoke, {
+        type: CryptoRequestType.GET_ADDRESS,
+        wallet
+      });
+
+      const uuid = generateDeterministicAddressUUID(address);
+
+      if (accounts.find((a) => a.uuid === uuid)) {
+        continue;
```
Mrtenz, line 101: Should we replace the previous entry, rather than ignoring it? For example if a user has a private key, but decides to use a mnemonic phrase later on?

Mrtenz (changes requested):

src/app/store/account.slice.ts
FrederikBolding: Good question, I don't know entirely what would be the logical way to handle it

Mrtenz (approved):

PR #58, opened 2021-03-12
Update onboarding flow

Mrtenz:


src/app/screens/__tests__/Login.test.tsx
```diff
@@ -56,18 +56,4 @@ describe('Login', () => {
 
     expect(mockStore.getActions()).toContainEqual(login('password'));
   });
-
-  it('dispatches login when pressing enter', async () => {
```
FrederikBolding, line 60: Why would we want to remove this test?

src/app/screens/CreatePassword.tsx
```diff
@@ -0,0 +1,95 @@
...
+const CREATE_PASSWORD_STRUCT = object({
+  password: refine(string(), 'Strong password', (value) => {
+    if (passwordStrength(value).id === PasswordStrength.STRONG) {
+      return true;
+    }
+
+    return translateRaw('PASSWORD_TOO_WEAK');
+  }),
+  passwordConfirmation: refine(string(), 'Equal to password', (value, context) => {
+    return value === context.branch[0].password;
```
FrederikBolding, line 34: This needs a better explainer text.

`Expected a value of type string, but received:"password1"`

src/app/screens/CreatePassword.tsx
```diff
@@ -0,0 +1,95 @@
...
+      password: '',
+      passwordConfirmation: ''
+    },
+    getValidator(CREATE_PASSWORD_STRUCT)
+  );
+
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    await form.validate();
```
FrederikBolding, line 51: Can we run form validation on input too? Perhaps debounced

src/app/screens/CreatePassword.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (form.error) {
+      return;
+    }
+
+    dispatch(createPassword(form.values.password));
+  };
+
+  return (
+    <Box>
+      <Heading fontSize="24px" lineHeight="150%" mt="24px" mb="8px" textAlign="center">
```
FrederikBolding, line 61: FYI you can use margin `1` `2` `3` etc. which uses theme values, that way it is standardized across the app

src/app/screens/CreatePassword.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <FormError name="passwordConfirmation" form={form} />
+        </Box>
+        <Flex mt="20px">
+          <Image src={warning} width="20px" height="20px" minWidth="20px" alt="Warning" mr="6px" />
+          <Body>
+            {translate('CREATE_PASSWORD_DESCRIPTION_2', {
+              $link: getKBHelpArticle(KB_HELP_ARTICLE.HOW_TO_BACKUP)
+            })}
+          </Body>
+        </Flex>
```
FrederikBolding, line 88: Box needs some margin to allow it to scroll under `PanelBottom`. Right now I can't see the help links at the bottom.

src/app/screens/Login.tsx
```diff
@@ -1,30 +1,27 @@
...
-      dispatch(login(password));
-    }
-  };
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
 
-  const handleKeyPress = (event: KeyboardEvent<HTMLInputElement>) => {
-    if (event.key === 'Enter') {
-      handleLogin();
+    if (form.error) {
```
FrederikBolding, line 19: How would there be an error without validation?

src/app/screens/__tests__/CreatePassword.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    fireEvent.change(passwordInput, { target: { value: RANDOM_PASSWORD } });
+
+    const createButton = getByText(translateRaw('CREATE_PASSWORD'));
+    expect(createButton).toBeDefined();
+    await waitFor(() => fireEvent.click(createButton));
+
+    expect(mockStore.getActions()).not.toContainEqual(createPassword(RANDOM_PASSWORD));
+  });
+
+  it('does not submit on weak passwords', async () => {
```
FrederikBolding, line 82: Can we test that validation messages are shown?

src/app/store/auth.slice.test.ts
```diff
@@ -78,6 +86,73 @@ describe('authSlice', () => {
...
+  });
+
+  it('dispatches createPasswordFailed on error', () => {
+    return expectSaga(createPasswordWorker, createPassword('foobar'))
+      .provide([[call.fn(ipcBridgeRenderer.db.invoke), false]])
+      .call(ipcBridgeRenderer.db.invoke, {
+        type: DBRequestType.INIT,
+        password: 'foobar'
+      })
+      .put(createPasswordFailed(translateRaw('LOGIN_ERROR')))
```
FrederikBolding, line 153: Can we use another string for this so it doesn't get confusing when we want to add better copy?

src/app/components/FormInput.tsx
```diff
@@ -0,0 +1,21 @@
...
+  Key extends keyof T,
+  Value extends T[Key] | T[Key][keyof T[Key]],
+  State extends DefaultState = DefaultState,
+  Error extends string = DefaultError
+>({
+  form,
+  children,
+  ...rest
+}: Omit<Omit<InputProps, 'form'> & FormInputProps<T, State, Error, Key, Value>, 'as'>) => (
+  <RebassInput as={ReactFormInput} form={form as any} {...rest}>
```
FrederikBolding, line 18: Can we prevent the `any` here?

FrederikBolding (changes requested):

src/app/screens/Login.tsx
Mrtenz: This is a placeholder for when we add validation to the login page. I can add it as part of this PR.

src/app/screens/__tests__/Login.test.tsx
Mrtenz: Because this is handled with the default browser functionality now, not with React handlers. Firing an event does not submit the form anymore.

src/app/components/FormInput.tsx
Mrtenz: Possibly, but I don't know how. The `@rebass/forms` input has a prop `form: string;` but the `typed-react-form` `FormInput` expects a `Form` object. Do you have suggestions for how to handle this?

src/app/screens/CreatePassword.tsx
Mrtenz: It would be better to handle this with flex, so the `PanelBottom` doesn't go over the content in the first place. I'll add some margin to fix it temporarily, and make a new PR for a more permanent solution.

src/app/components/FormInput.tsx
FrederikBolding: What is it used for exactly?

src/app/screens/Login.tsx
FrederikBolding: 👍

src/app/screens/CreatePassword.tsx
FrederikBolding: Sure

src/app/screens/Login.tsx
```diff
@@ -1,30 +1,40 @@
...
+      return true;
+    }
+
+    return translateRaw('PASSWORD_EMPTY');
+  })
+});
+
 export const Login = () => {
-  const [password, setPassword] = useState('');
+  const form = useForm({ password: '' }, getValidator(LOGIN_STRUCT), true, false);
```
FrederikBolding, line 24: Last param is the default value right?

FrederikBolding (approved):

PR #61, opened 2021-03-18
More add account flow

FrederikBolding:


src/app/components/Core/Selector.tsx
```diff
@@ -0,0 +1,208 @@
...
+  isClearable?: boolean;
+  components?: SelectComponentsConfig<T>;
+  getOptionLabel?(option: T): string;
+  onChange?(option: T): void;
+  onInputChange?(newValue: string, actionMeta: InputActionMeta): void;
+  onInputKeyDown?(e: React.KeyboardEvent<HTMLElement>): void;
+}
+
+// Fixes weird placement issues for react-select
+const Wrapper = styled('div')`
```
Mrtenz, line 44:
```suggestion
const Wrapper = styled.div`
```

src/app/components/Core/Selector.tsx
```diff
@@ -0,0 +1,208 @@
...
+  options: T[];
+  // We prefer controlled components so `value` is required prop. When it is `null`, React-Select will display the placeholder
+  value: T | null | undefined;
+  disabled?: boolean;
+  placeholder?: string;
+  searchable?: boolean;
+  clearable?: boolean;
+  name?: string;
+  optionComponent: React.ComponentType<OptionProps<T>>;
+  valueComponent?: React.ComponentClass<{ value: T }> | React.StatelessComponent<{ value: T }>;
```
Mrtenz, line 28: `StatelessComponent` is deprecated, you can use `FunctionComponent` instead.

src/app/components/Core/Selector.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  return (
+    <ReactSelectComponents.ValueContainer {...props}>
+      {React.Children.map(children, (child: JSX.Element) =>
+        child && [ReactSelectComponents.SingleValue].indexOf(child.type) === -1 ? child : null
+      )}
+      {!inputValue && <ValueComponent value={data} />}
+    </ReactSelectComponents.ValueContainer>
+  );
+};
```
Mrtenz, line 95: This function can be typed a lot better, e.g.:
```suggestion
const getValueContainer = <T extends unknown = unknown>(
  props: ValueContainerProps<T> & OptionProps<T>,
  ValueComponent?: ComponentType<{ value: T }>
): ReactElement => {
  const {
    hasValue,
    getValue,
    selectProps: { inputValue }
  } = props;

  if (!hasValue || !ValueComponent) {
    return <ReactSelectComponents.ValueContainer {...props} />;
  }

  const [data] = getValue() as T[];
  const { children } = props;

  return (
    <ReactSelectComponents.ValueContainer {...props}>
      {React.Children.map(children, (child: JSX.Element) =>
        child && [ReactSelectComponents.SingleValue].indexOf(child.type) === -1 ? child : null
      )}
      {!inputValue && <ValueComponent value={data} />}
    </ReactSelectComponents.ValueContainer>
  );
};
```

src/app/components/Core/Selector.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+const getOption = (
+  { optionDivider = false, ...props }: OptionProps<any> & { optionDivider: boolean },
+  Component: React.ComponentType<OptionProps<any>>
```
Mrtenz, line 99: Use generics to replace `any`.

src/app/components/Core/Selector.tsx
```diff
@@ -0,0 +1,208 @@
...
+      components={{
+        DropdownIndicator: (props) => {
+          return (
+            <DropdownIndicatorWrapper {...props}>
+              <Image src={expand} height="6px" />
+            </DropdownIndicatorWrapper>
+          );
+        },
+        Option: (oProps: any) => getOption({ ...oProps, optionDivider }, optionComponent),
+        ValueContainer: (oProps: any) => getValueContainer(oProps, valueComponent),
```
Mrtenz, line 199: Use generics to replace `any`.

src/app/components/FileBox.tsx
```diff
@@ -1,45 +1,81 @@
...
-    //onDrop={() => console.log('drop')}
-  >
+export const FileBox = ({
+  onChange,
+  ...props
+}: { onChange(file: File): void } & Omit<BoxProps, 'onChange'>) => {
+  const [fileName, setFileName] = useState<string | undefined>(undefined);
+
+  // Needs to be set for onDrop to work
+  const handleDragOver = (e: React.DragEvent) => {
```
Mrtenz, line 14: Something to consider: The new JSX transform will remove the need for `React` imports. It may be more future-proof to import directly, e.g. `import { DragEvent } from 'react';`, since then we can easily remove all `React` imports if we deccide to switch to the JSX transform. Then the imports from React are essentially type-only imports.

Mrtenz (changes requested):

FrederikBolding:
[ch8259]

Mrtenz (approved):

PR #72, opened 2021-03-31
Refactor panel to use flexbox and fix consistency issues

Mrtenz:


src/app/screens/AddAccount/AddAccount.tsx
```diff
@@ -18,7 +18,9 @@ export const AddAccount = () => {
 
   return (
     <>
-      <WalletTypeSelector walletType={walletType} setWalletType={setWalletType} />
+      <Container sx={{ flex: 'none' }}>
```
FrederikBolding, line 21: Everything is supposed to scroll here, so we should stick to one `Container`

src/app/components/Container.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { Box, BoxProps } from '@components';
+
+export const Container = ({ children, ...props }: BoxProps) => {
+  return (
+    <Box
+      pt="4"
```
FrederikBolding: Double the required padding than designs

src/app/components/SignBottom.tsx
```diff
@@ -6,14 +6,16 @@ import { translateRaw } from '@translations';
...
+  form
 }: {
   disabled: boolean;
   handleAccept?(): void;
   handleDeny(): void;
+  form?: string;
 }) => (
   <PanelBottom>
-    <Button id="accept_button" type="submit" disabled={disabled} onClick={handleAccept}>
+    <Button id="accept_button" type="submit" form={form} disabled={disabled} onClick={handleAccept}>
```
FrederikBolding, line 18: Why?

src/app/screens/GenerateAccount/GenerateAccountStart.tsx
```diff
@@ -1,26 +1,38 @@
...
+      <List>
+        <ListItem>{translateRaw('MNEMONIC_DESCRIPTION_1')}</ListItem>
+        <ListItem>{translateRaw('MNEMONIC_DESCRIPTION_2')}</ListItem>
+        <ListItem>{translateRaw('MNEMONIC_DESCRIPTION_3')}</ListItem>
+        <ListItem>{translateRaw('MNEMONIC_DESCRIPTION_4')}</ListItem>
+        <ListItem>{translateRaw('MNEMONIC_DESCRIPTION_5')}</ListItem>
+      </List>
+    </Container>
+    <PanelBottom variant="clear">
+      <Button onClick={onNext}>{translateRaw('CREATE_MNEMONIC_PHRASE')}</Button>
```
FrederikBolding, line 35: This button looks off

src/app/screens/GenerateAccount/GenerateAccountVerify.tsx
```diff
@@ -41,45 +51,49 @@ export const GenerateAccountVerify = ({ onNext }: IFlowComponentProps) => {
...
+              </Body>
+            </Label>
+            <FormInput id="thirdWord" name="thirdWord" data-testid="thirdWord" form={form} />
+            <FormError name="thirdWord" form={form} />
+          </Box>
+        </form>
+      </Container>
+      <PanelBottom variant="clear">
+        <Button type="submit" form="mnemonic-phrase-verify-form">
           {translateRaw('VERIFY_WORDS')}
```
FrederikBolding, line 103: This button looks off - not centered at the bottom. Is this a pattern we want to use instead of the ones with the white background? Those look better imo

src/app/screens/Login.tsx
```diff
@@ -48,26 +49,30 @@ export const Login = () => {
...
-      <Box>
-        <Body mt="16px">
+        {/* @todo: Pretty error messages */}
+        {error}
+      </Container>
+      <PanelBottom variant="clear">
+        <Button type="submit" form="login-form">
+          {translateRaw('UNLOCK_NOW')}
+        </Button>
+        <Body mt="3">
```
FrederikBolding, line 68: Not centered text

src/app/screens/Login.tsx
```diff
@@ -48,26 +49,30 @@ export const Login = () => {
...
             <FormError name="password" form={form} />
           </Box>
-          <Button mt="24px" type="submit">
-            {translateRaw('UNLOCK_NOW')}
-          </Button>
         </form>
-      </Box>
-      <Box>
-        <Body mt="16px">
+        {/* @todo: Pretty error messages */}
```
FrederikBolding, line 61: Just use `Body` with `error` variant

src/app/screens/GenerateAccount/GenerateAccountStart.tsx
FrederikBolding: It's because it doesn't have a background and is not centered, it looks weird when we have scrolling

src/app/screens/SignTransaction/SignTransaction.tsx
```diff
@@ -48,13 +48,15 @@ export const SignTransaction = () => {
 
   return (
     <>
-      <Box variant="rowAlign" mb="2">
-        <Blockie height="32px" width="32px" address={currentAccount.address} mr="2" />
-        <Box>
-          <Body>{translateRaw('NO_LABEL')}</Body>
-          <Body fontSize="14px">{currentAccount.address}</Body>
+      <Container sx={{ flex: 'none' }}>
```
FrederikBolding, line 51: This part should be scrollable too

src/app/components/Flow/Flow.tsx
```diff
@@ -37,13 +40,14 @@ export const Flow = ({ components, onDone }: FlowProps) => {
...
   return (
     <>
-      <FlowHeader
-        onPrevious={handlePrevious}
-        steps={components.length}
-        currentStep={step}
-        mt="3"
-        mb="2"
-      />
+      <Container pt="4" sx={{ flex: 'none' }}>
```
FrederikBolding, line 43: I think there is too much padding here too, even though the designs look similar.

FrederikBolding (changes requested):

src/app/components/SignBottom.tsx
Mrtenz: This makes the button submit `form`, even though it´s not directly inside that form

FrederikBolding (approved):

PR #75, opened 2021-04-05
Account Management + Labels

FrederikBolding:


FrederikBolding:
[ch8438]

src/app/components/EditableText.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const hasValue = value !== undefined;
+
+  return (
+    <Box variant="rowAlign" height="100%" {...props}>
+      {editMode ? (
+        <Input
+          placeholder={placeholder}
+          autoFocus={true}
+          value={editValue}
+          onChange={(e) => setEditValue(e.currentTarget.value)}
```
Mrtenz, line 56: Don´t use an arrow function here

src/app/components/EditableText.tsx
```diff
@@ -0,0 +1,95 @@
...
+    onChange(editValue);
+    setEditMode(false);
+  };
+
+  const hasValue = value !== undefined;
+
+  return (
+    <Box variant="rowAlign" height="100%" {...props}>
+      {editMode ? (
+        <Input
```
Mrtenz, line 52: This input is being changed from uncontrolled to controlled when writing something in it. I´m not sure why, but this should be avoided if possible.

Mrtenz (changes requested):
Few things regarding the design:
- I think there should be slightly more padding between the blocky/delete icon and address
- The blockie should be top aligned
- The font size is too big for the address
- The delete overlay should span the entire width

Mrtenz (approved):

PR #76, opened 2021-04-07
Packaging workflow

GuillaumeRx:
This PR adds :
- Building and uploading of artifacts on PRs
- Building and publishing release artifacts on tag
- Notarisation of MacOs build
- Windows signing step (not active)


GuillaumeRx:
[ch8415]

entitlements.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+    <dict>
+        <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
```
FrederikBolding, line 5: Do we need this? If so, why?

forge.config.js
```diff
@@ -0,0 +1,71 @@
+module.exports = {
+  packagerConfig: {
+    osxSign: {
+      identity: 'Developer ID Application: MyCrypto Inc (43RN6VJTB7)',
+      'hardened-runtime': true,
+      'gatekeeper-assess': false,
```
FrederikBolding, line 6: Why is this disabled?

.github/workflows/package.yml
```diff
@@ -0,0 +1,99 @@
...
+      # To add when we have a Windows signing certificate
+      # - name: Add Windows certificate
+      #   if: matrix.os == 'windows-latest' && startsWith(github.ref, 'refs/tags/')
+      #   id: write_file
+      #   uses: timheuer/base64-to-file@v1
+      #   with:
+      #     fileName: 'win-certificate.pfx'
+      #     encodedString: ${{ secrets.CERTIFICATE_WINDOWS_PFX }}
+
+      - name: Make artifact (${{ matrix.os }})
```
FrederikBolding, line 70: This gives the following warning: `You have set packagerConfig.ignore, the Electron Forge webpack plugin normally sets this automatically. Your packaged app may be larger than expected if you dont ignore everything other than the '.webpack' folder`

I don't see changes to this though, but I haven't gotten that error locally.

FrederikBolding (changes requested):
Looking good! A few small questions and the PR needs a rebase 😄

.github/workflows/package.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-build-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: yarn install
+
+      - name: Add MacOS certificates
+        if: matrix.os == 'macos-latest'
```
FrederikBolding, line 44: Can we skip the MacOS signing when doing pull request builds? There is no reason to sign the builds and it might be faster to build without?

FrederikBolding (changes requested):

entitlements.plist
GuillaumeRx: This is the minimal requirement to pass notarisation with electron :) If we don't use that it will not succeed

forge.config.js
GuillaumeRx: This also needs to be disable since we use self signed certificates

.github/workflows/package.yml
GuillaumeRx: Yeah it showed up when I moved the forge config outside of the package.json. Will look into that

.github/workflows/package.yml
GuillaumeRx: There is no flags available to skip signing :(

Mrtenz: This step should only be necessary for releases right?

.github/workflows/package.yml
```diff
@@ -0,0 +1,99 @@
...
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest, windows-latest]
+    steps:
+      - name: Retrieves project from the git branch
+        uses: actions/checkout@v2
+
+      - name: Read .nvmrc
+        run: echo "##[set-output name=NVMRC;]$(cat package.json | grep '\"node\":' | sed 's/^ *//;s/ *$//;s/\"node\":\ \"//;s/\",//' )"
+        id: nvm
```
Mrtenz, line 23: This command doesn't seem to read anything from .nvmrc, at least from my understanding of it. We can just hardcode a Node version like we did in the CI workflow: https://github.com/MyCryptoHQ/signer/blob/master/.github/workflows/ci.yml#L17-L18

.github/workflows/package.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Cache node modules
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-build-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
```
Mrtenz, line 40: This is redundant.

.github/workflows/package.yml
```diff
@@ -0,0 +1,99 @@
...
+        run: yarn install
```
Mrtenz, line 41: Please add `--frozen-lockfile`.

.github/workflows/package.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          name: ${{ matrix.os }} build
+          path: |
+            out/make/**/*.exe
+            out/make/**/*.rpm
+            out/make/**/*.deb
+            out/make/**/*.zip
+            out/make/**/*.nupkg
+
+      - name: Make and release (${{ matrix.os }})
```
Mrtenz, line 91: The `matrix.os` here (and in some other steps) shouldn't be necessary right? Since GitHub already separates the runs per OS?

entitlements.plist
Mrtenz: The Apple docs mention:
> In rare cases, an app might need to override or patch C code, use the long-deprecated NSCreateObjectFileImageFromMemory (which is fundamentally insecure), or use the DVDPlayback framework. Add the Allow Unsigned Executable Memory Entitlement to enable these use cases. Otherwise, the app might crash or behave in unexpected ways.
> ([Source](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-unsigned-executable-memory))

Is it safe to set this flag?

forge.config.js
```diff
@@ -0,0 +1,71 @@
...
+      entitlements: 'entitlements.plist',
+      'entitlements-inherit': 'entitlements.plist',
+      'signature-flags': 'library'
+    },
+    osxNotarize: {
+      appleId: process.env['APPLE_ID'],
```
Mrtenz, line 12:
```suggestion
      appleId: process.env.APPLE_ID,
```

forge.config.js
```diff
@@ -0,0 +1,71 @@
...
+      appleIdPassword: process.env['APPLE_ID_PASSWORD']
```
Mrtenz, line 13:
```suggestion
      appleIdPassword: process.env.APPLE_ID_PASSWORD
```

Mrtenz (changes requested):

entitlements.plist
GuillaumeRx: It's as safe as Electron is 😅 I'll try to remove it to see if it still work 😄

.github/workflows/package.yml
GuillaumeRx: Well, it's also nice to be able to open any builds vithout having to fight with macOs IMO and if we want to skip signing we would need two configs. There is no config options/flags to avoir this step during build as soon as you provided the config to sign

.github/workflows/package.yml
GuillaumeRx: You're right. Took it from the CI workflow before its rework - Removing :)

.github/workflows/package.yml
GuillaumeRx: Nope, it's just for display in github :) I can remove if you prefer

.github/workflows/package.yml
```diff
@@ -13,19 +13,17 @@ jobs:
...
         os: [macos-latest, ubuntu-latest, windows-latest]
     steps:
       - name: Retrieves project from the git branch
         uses: actions/checkout@v2
 
-      - name: Read .nvmrc
-        run: echo "##[set-output name=NVMRC;]$(cat package.json | grep '\"node\":' | sed 's/^ *//;s/ *$//;s/\"node\":\ \"//;s/\",//' )"
-        id: nvm
-
       - name: Setup node ${{ steps.nvm.outputs.NVMRC }}
```
Mrtenz, line 23: Make sure to remove `${{ steps.nvm.outputs.NVMRC }}` from here

entitlements.plist
GuillaumeRx: Well instantly crashed :(

FrederikBolding (approved):
lgtm - please wait with merging until @Mrtenz has approved it too.

Mrtenz (approved):

PR #77, opened 2021-04-13
Use Redux as synchronisation layer for the different Electron processes

Mrtenz:


src/bridge.ts
```diff
@@ -41,10 +41,28 @@ const getChannel = <A, B>(channel: IPC_CHANNELS) => {
...
+        ipcRenderer.removeListener(channel, listener);
+      };
+    },
     invoke: (request: A) => ipcRenderer.invoke(channel, request)
   });
 
-  const asMain = (ipcMain: IpcMain) => ({
+  const asMain = (ipcMain: IpcMain, Contents: typeof WebContents) => ({
+    emit: (...args: unknown[]) => {
+      Contents.getAllWebContents().forEach((webContents) => webContents.send(channel, ...args));
```
FrederikBolding, line 57: I would prefer using the one exposed by main only. This could be a security issue.

src/api/store/store.ts
```diff
@@ -0,0 +1,33 @@
...
+import { synchronisationMiddleware } from '@common/store';
+
+import { createRootReducer } from './reducer';
+import rootSaga from './sagas';
+
+const reducer = createRootReducer();
+
+export type ApplicationState = ReturnType<typeof reducer>;
+
+export const createStore = (): EnhancedStore<ApplicationState> => {
```
FrederikBolding, line 15: Is it possible to get coverage for this?

src/common/store/handshake.ts
```diff
@@ -0,0 +1,152 @@
...
+      }
+    });
+
+    return () => {
+      unsubcribe();
+    };
+  });
+};
+
+export function* putJson(json: string, allowInsecure: boolean = false): SagaIterator {
```
FrederikBolding, line 102: Can we just call `allowInsecure` `isDecrypted` or something like that? May be clearer

src/common/store/synchronisation.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ReduxIPC } from '@types';
+
+/**
+ * Middleware that dispatches any actions to the other Electron process.
+ * @param ipc The Electron process to dispatch from.
+ */
+export const synchronisationMiddleware = (ipc: ReduxIPC): Middleware => (store) => (next) => (
+  action
+) => {
+  if (
```
FrederikBolding, line 14: Should we have a better way of ignoring sensitive action types here? In the future we might want to ignore other things right?

src/translations/translateRaw.tsx
```diff
@@ -1,77 +1,6 @@
...
-      str = str.replace(re, variables[variable]);
-    });
-
-    return str;
-  }
-  return translatedString;
-}
+import { fallbackLanguage, repository } from '@common/translate';
 
 export const Trans = ({
```
FrederikBolding, line 5: Rename this file

src/common/store/handshake.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      .silentRun();
+
+    const insecureAction = JSON.stringify(setHandshaken(true));
+    await expectSaga(putJson, insecureAction, true)
+      .put({ ...setHandshaken(true), remote: true })
+      .silentRun();
+  });
+
+  it('decrypts an encrypted action', async () => {
+    const insecureAction = JSON.stringify(setHandshaken(true));
```
FrederikBolding, line 76: Can we test with something that is supposed to be synced? This is a bit asking for trouble 😆

FrederikBolding: Same goes for the other tests in this file

src/common/store/handshake.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  it('generates a keypair and dispatches setKeyPair', async () => {
+    const keyPair = await createHandshakeKeyPair();
+
+    await expectSaga(createKeyPairWorker, createKeyPair()).put(setKeyPair(keyPair)).silentRun();
+  });
+
+  it('sends the public key if the payload is true', async () => {
+    const keyPair = await createHandshakeKeyPair();
+
+    await expectSaga(createKeyPairWorker, createKeyPair(true))
```
FrederikBolding, line 102: Can we document this somewhere in the code that the renderer initializes the handshake process with this call?

src/common/store/handshake.ts
```diff
@@ -0,0 +1,152 @@
...
+  yield all([
+    ipcWorker(ipc),
+    takeLatest(createKeyPair.type, createKeyPairWorker),
+    takeEvery(sendPublicKey.type, setPublicKeyWorker)
+  ]);
+}
+
+export const subscribe = (ipc: ReduxIPC) => {
+  return eventChannel((emitter) => {
+    const unsubcribe = ipc.on((_: Event, action: string) => {
```
FrederikBolding, line 90: Could this be abstracted to a helper that can be used in the transactions saga too?

FrederikBolding (changes requested):

src/common/store/handshake.ts
```diff
@@ -0,0 +1,150 @@
...
+  isHandshaken: boolean;
+
+  targetPublicKey?: string;
+}
+
+const initialState: HandshakeState = {
+  isHandshaken: false
+};
+
+const sliceName = 'handshake';
```
FrederikBolding, line 29: Lets rename this slice and sagas to synchronization for clarity

src/common/store/synchronisation.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * An array of action paths that will not be synchronised with the other process.
+ */
+export const IGNORED_PATHS = [handshake.name];
+
+/**
+ * Middleware that dispatches any actions to the other Electron process.
+ * @param ipc The Electron process to dispatch from.
+ */
+export const synchronisationMiddleware = (ipc: ReduxIPC): Middleware => (store) => (next) => (
```
FrederikBolding, line 20: Rename file to make it clear that this file is the middleware

FrederikBolding (changes requested):

FrederikBolding (approved):

FrederikBolding:
[ch8469]

PR #83, opened 2021-04-19
Crypto Process

FrederikBolding:


FrederikBolding:
[ch8497]

src/api/store/accounts.sagas.test.ts
```diff
@@ -1,33 +1,17 @@
-import type { TAddress } from '@mycrypto/wallets';
-import { DEFAULT_ETH } from '@mycrypto/wallets';
+// @fixme DONT DO THIS
```
Mrtenz, line 1: What´s the reason for this?

src/api/store/accounts.sagas.ts
```diff
@@ -1,43 +1,33 @@
-import { DEFAULT_ETH } from '@mycrypto/wallets';
 import type { PayloadAction } from '@reduxjs/toolkit';
+// @fixme DONT DO THIS
```
Mrtenz, line 2: Same as above.

src/common/store/signing.slice.ts
```diff
@@ -13,6 +13,9 @@ const slice = createSlice({
   name: sliceName,
   initialState,
   reducers: {
+    init(_, __: PayloadAction<string>) {
+      // noop
+    },
```
Mrtenz, line 18: This can be a separate action, if it does not modify state.

src/api/store/sagas.ts
```diff
@@ -1,21 +1,19 @@
...
 import type { ReduxIPC } from '@types';
 
 import { accountsSaga } from './accounts.sagas';
 import { authSaga } from './auth.sagas';
-import { signingSaga } from './signing.sagas';
 import { transactionsSaga } from './transactions.sagas';
 import { webSocketSaga } from './ws.sagas';
 
-export default function* rootSaga(ipc: ReduxIPC) {
+export default function* rootSaga(ipcs: Partial<Record<SynchronizationTarget, ReduxIPC>>) {
```
Mrtenz, line 11: Maybe name this `processes` instead of `ipcs`.

src/common/store/synchronization.middleware.ts
```diff
@@ -1,56 +1,100 @@
...
+      ? SynchronizationTarget.SIGNING
+      : SynchronizationTarget.MAIN
+  };
+
+  const target = ipcMapping[self];
+
+  const encryptionTarget = encryptionMapping[self];
+
+  // UNDEFINED = ALL
+  const to = action.to ?? SIGNING_ACTIONS.includes(action.type) ? encryptionTarget : undefined;
```
Mrtenz, line 57: Maybe add an `All` value to the enum. That´s a bit more clear than using `undefined` here.

src/common/store/synchronization.slice.ts
```diff
@@ -15,17 +15,24 @@ import {
 import type { HandshakeKeyPair, ReduxIPC } from '@types';
 import { safeJSONParse } from '@utils';
 
+export enum SynchronizationTarget {
+  RENDERER = 'RENDERER',
+  MAIN = 'MAIN',
+  SIGNING = 'SIGNING'
+}
```
Mrtenz, line 22: Maybe name this `Process` instead of `SynchronizationTarget`. Also, maybe we should follow [TypeScript´s naming conventions for enums](https://www.typescriptlang.org/docs/handbook/enums.html), e.g., `Renderer`, `Main`, `Signing`.

src/common/store/synchronization.slice.ts
```diff
@@ -15,17 +15,24 @@ import {
...
+
 interface SynchronizationState {
   publicKey?: string;
   privateKey?: string;
 
-  isHandshaken: boolean;
+  isHandshaken: Partial<Record<SynchronizationTarget, boolean>>;
```
Mrtenz, line 28: Why is this `Partial`? Can´t it be an object with the value set to `false` by default?

src/common/store/synchronization.slice.ts
```diff
@@ -70,18 +83,20 @@ export const getSynchronizationState = createSelector(
...
 export const postHandshake = createAction(`${sliceName}/postHandshake`);
 
-export function* handshakeSaga(ipc: ReduxIPC) {
+export function* handshakeSaga(
+  ipcs: Partial<Record<SynchronizationTarget, ReduxIPC>>,
+  self: SynchronizationTarget
+) {
   yield all([
-    ipcWorker(ipc),
+    ...Object.keys(ipcs).map((target) => ipcWorker(ipcs, target as SynchronizationTarget, self)),
```
Mrtenz, line 99: What´s the reason for the type cast here? Is TypeScript not able to infer the type?

src/common/store/synchronization.slice.ts
```diff
@@ -99,31 +114,53 @@ export const subscribe = (ipc: ReduxIPC) => {
...
+  console.log(self, 'Received', action);
+
   if (isReduxAction(action) && (isDecrypted || action.type === sendPublicKey.type)) {
     yield put({ ...action, remote: true });
     return;
   }
 
-  const isHandshaken = yield select(getHandshaken);
+  const from = action.from as SynchronizationTarget | undefined;
+  const to = action.to as SynchronizationTarget | undefined;
```
Mrtenz, line 136: What´s the reason for the type cast here?

src/common/utils/synchronization.ts
```diff
@@ -11,7 +11,9 @@ const REDUX_ACTION_STRUCT = type({
 });
 
 const ENCRYPTED_ACTION_STRUCT = object({
-  data: string()
+  data: string(),
+  to: optional(string()),
```
Mrtenz, line 15: Should this be optional? What´s the behaviour when `to` is not specified?

src/main.ts
```diff
@@ -60,7 +62,22 @@ const createWindow = (): void => {
...
+    [SynchronizationTarget.SIGNING]: {
+      on: (listener: (event: IpcMainEvent, ...args: unknown[]) => void) => {
+        const callback = (msg: unknown) => listener(undefined, msg);
+        signingProcess.on('message', callback);
+        return () => {
+          signingProcess.removeListener('message', callback);
+        };
+      },
+      emit: (args) => signingProcess.send(args as string)
+    }
```
Mrtenz, line 78: Can we move this logic to a separate file and add tests?

webpack_config/main.js
```diff
@@ -5,7 +5,7 @@ const { merge } = require('webpack-merge');
 const common = require('./common');
 
 module.exports = merge(common, {
-  entry: './src/main.ts',
+  entry: { index: './src/main.ts', worker: './src/signing/signing.worker.ts' },
```
Mrtenz, line 8: We may need a separate Webpack config for the worker, to make sure it does not include any unnecessary dependencies and stuff.

Mrtenz (changes requested):

src/common/store/synchronization.middleware.ts
FrederikBolding: I did this initially, but then it would require the mappings above to include `All` to not be a typing mess. But that doesn't really make sense

src/common/store/synchronization.slice.ts
FrederikBolding: No, it just assumes it is a string

src/common/utils/synchronization.ts
FrederikBolding: Then it is broadcast to all

src/common/store/synchronization.slice.ts
FrederikBolding: I guess I could type it in the `safeJSONParse`. Would you prefer that?

src/api/store/accounts.sagas.test.ts
FrederikBolding: We probably want to move this to the crypto process since it touches private keys. Just haven't moved it yet

webpack_config/main.js
FrederikBolding: Yeah, probably. Not sure how to go about that for right now. But we should do it before merging. Maybe we can talk about this later?

src/common/store/synchronization.slice.ts
FrederikBolding: It can, but is there a reason to store your own value that'll always be `false`? Since you can't handshake with yourself

src/common/store/accounts.slice.ts
```diff
@@ -40,7 +40,7 @@ const slice = createSlice({
     },
     removeAccount(state, action: PayloadAction<IAccount>) {
       const idx = state.accounts.findIndex((a) => a.uuid === action.payload.uuid);
-      state.accounts.splice(idx, 1);
+      if (idx != -1) state.accounts.splice(idx, 1);
```
Mrtenz, line 43: We shouldn't use if statements in reducers.

src/crypto/crypto.worker.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { createStore } from './store';
+
+export const init = () => {
+  const store = createStore({
+    emit: (msg) => {
+      console.log('SIGNING EMITTING', msg);
+      process.send(msg);
+    },
+    on: (listener: any) => {
```
Mrtenz, line 11: This can be properly typed.

src/crypto/crypto.worker.ts
```diff
@@ -0,0 +1,23 @@
...
+      process.on('message', (msg) => listener(undefined, msg));
+      process.on('message', (msg) => console.log('SIGNING RECEIVED', msg));
+      return () => {
+        //process.removeListener('message', listener);
```
Mrtenz, line 15: Why is this commented out?

src/ipc.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Process } from '@common/store';
+
+jest.unmock('@bridge');
+
+const mockWindow = { webContents: { send: jest.fn() } };
+const mockCrypto = { send: jest.fn(), on: jest.fn(), removeListener: jest.fn() };
+
+describe('createIpc', () => {
+  it('emits to browser window when targeting Renderer', () => {
+    const ipc = createIpc((mockWindow as unknown) as BrowserWindow, mockCrypto as any);
```
Mrtenz, line 14: Maybe use something like `mockCrypto as unknown as ReturnType<typeof createCryptoProcess>` to avoid `any`.

Mrtenz (changes requested):

src/common/store/synchronization.middleware.ts
```diff
@@ -1,54 +1,92 @@
 import type { Middleware } from '@reduxjs/toolkit';
 import { PERSIST } from 'redux-persist';
 
+import { init, sign } from '@common/store/signing.slice';
```
Mrtenz, line 4: Use relative import.

src/common/store/synchronization.middleware.ts
```diff
@@ -1,54 +1,92 @@
...
   if (
     (action.type !== sendPublicKey.type && IGNORED_PATHS.includes(path)) ||
     IGNORED_ACTIONS.includes(action.type) ||
-    action.remote
+    (action.remote && target === action.from) ||
+    to === self ||
+    (from !== self && self !== Process.Main)
   ) {
     return next(action);
   }
```
Mrtenz, line 68: Maybe move this into a separate function and split into multiple if statements, to make it more readable.

src/common/store/synchronization.slice.ts
```diff
@@ -40,15 +47,15 @@ const slice = createSlice({
     sendPublicKey(_, __: PayloadAction<string>) {
       // noop
     },
```
Mrtenz, line 49: Move to separate action.

src/common/store/synchronization.slice.ts
```diff
@@ -70,18 +77,17 @@ export const getSynchronizationState = createSelector(
...
+export const getTargetPublicKey = (target: Process) =>
+  createSelector(getSynchronizationState, (state) => state.targetPublicKey[target]);
 
 export const postHandshake = createAction(`${sliceName}/postHandshake`);
 
-export function* handshakeSaga(ipc: ReduxIPC) {
+export function* handshakeSaga(processes: Partial<Record<Process, ReduxIPC>>, self: Process) {
   yield all([
-    ipcWorker(ipc),
+    ...Object.keys(processes).map((target) => ipcWorker(processes, target as Process, self)),
```
Mrtenz, line 90: For better type safety, maybe we can use a wrapper function of Object.keys with proper types.
```ts
type ObjectKeys<T> =
  T extends Record<string, unknown> ? (keyof T)[] :
    T extends number ? [] :
      T extends Array<any> | string ? string[] :
        never;

const keys = <T extends Record<string, unknown>>(object: T): ObjectKeys<T> => {
  return Object.keys(object) as ObjectKeys<T>;
}
```

src/crypto/crypto.worker.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { createStore } from './store';
+
+export const init = () => {
+  const store = createStore({
+    emit: (msg) => process.send(msg),
+    on: (listener: (event: undefined, ...args: unknown[]) => void) => {
+      process.on('message', (msg) => listener(undefined, msg));
+      return () => {
+        (process as NodeJS.EventEmitter).removeListener('message', listener);
```
Mrtenz, line 11: Can this be tested?

webpack_config/main.js
```diff
@@ -5,7 +5,7 @@ const { merge } = require('webpack-merge');
 const common = require('./common');
 
 module.exports = merge(common, {
-  entry: './src/main.ts',
+  entry: { index: './src/main.ts', worker: './src/crypto/crypto.worker.ts' },
```
Mrtenz, line 8: Remove `worker` from main config.

Mrtenz (changes requested):

src/crypto/crypto.worker.ts
FrederikBolding: Not really unfortunately

Mrtenz (approved):

PR #94, opened 2021-04-29
Use separate key for encrypting settings

Mrtenz:


Mrtenz:
[ch8542]

src/utils/encryption.ts
```diff
@@ -13,6 +13,10 @@ const ENCRYPTION_IV_LENGTH = 12;
 const ENCRYPTION_TAG_LENGTH = 16;
 const ENCRYPTION_ALGORITHM = 'aes-256-gcm';
 
+export const createEncryptionKey = (size: number = KEY_LENGTH): Buffer => {
```
FrederikBolding, line 16: Can we mark that this is only to be used for settings, maybe with a comment?

src/common/store/storage/persistence.middleware.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      }
+    });
+
+    persistenceMiddleware()(store)(next)(rehydrateAllState());
+
+    expect(next).toHaveBeenCalled();
+    expect(store.getActions()).toStrictEqual([fetchSettings('foo'), fetchSettings('bar')]);
+  });
+
+  it('persists the whitelisted state if a whitelisted action is dispatched', () => {
```
FrederikBolding, line 37: Can we test the failure state here?

src/crypto/secrets.ts
```diff
@@ -40,3 +48,35 @@ export const saveAccountSecrets = async (initialiseWallet: SerializedWallet) =>
...
+};
+
+export const hasSettingsKey = async (): Promise<boolean> => {
+  const key = await keytar.getPassword(KEYTAR_SERVICE, KEYTAR_SETTINGS_KEY_NAME);
+  return !!key;
+};
+
+export const checkSettingsKey = async (): Promise<boolean> => {
+  try {
+    await getSettingsKey();
```
FrederikBolding, line 77: I'm worried this will potentially cause a weird bug where we accidentally create a settings key when we don't really want to and possibly have a false positive.

src/crypto/secrets.test.ts
```diff
@@ -66,3 +74,68 @@ describe('deleteAccountSecrets', () => {
...
+      .mockImplementationOnce(async () => 'foo')
+      .mockImplementationOnce(async () => undefined);
+
+    await expect(hasSettingsKey()).resolves.toBe(true);
+    await expect(hasSettingsKey()).resolves.toBe(false);
+  });
+});
+
+describe('checkSettingsKey', () => {
+  it('checks if the settings key can be decrypted', async () => {
```
FrederikBolding, line 129: Can we test some of the error states here?

src/crypto/auth.sagas.ts
```diff
@@ -1,14 +1,77 @@
...
+  }
+
+  yield put(loginFailed(translateRaw('LOGIN_ERROR')));
+}
+
+export function* createPasswordWorker({ payload }: PayloadAction<string>) {
+  yield call(resetWorker);
+
+  yield call(init, payload);
+  yield call(getSettingsKey);
```
FrederikBolding, line 63: Should we potentially mirror `loginWorker` here and use `checkSettingsKey`? Then we can catch potential errors?

src/crypto/auth.sagas.ts
```diff
@@ -1,14 +1,77 @@
...
+}
+
+export function* logoutWorker() {
+  yield call(clearEncryptionKey);
 }
 
 export function* resetWorker() {
+  yield put(resetSettings());
+
   const credentials = yield call(keytar.findCredentials, KEYTAR_SERVICE);
```
FrederikBolding, line 75: This should wipe the settings key too right? We need to make sure that is the case, otherwise you will be stuck when resetting.

src/app/screens/__tests__/Home.test.tsx
```diff
@@ -33,7 +33,7 @@ const mockStore = createMockStore({
     history: [historyTx, historyTx]
   },
   persistence: {
-    isPersisted: true
+    rehydratedKeys: []
```
FrederikBolding, line 36: How come this can be empty? Shouldn't it require all keys to be there to be an equivalent state to the old `isPersisted=true`

src/common/store/storage/reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface PersistState {
+  _persistence?: {
+    whitelistedActions: string[];
+    whitelistedKeys: string[];
+  };
+}
+
+export const injectPersistent = <A>(
+  { key, ...config }: PersistConfig,
```
FrederikBolding, line 20: If we filter out the key here do we actually need it elsewhere or can it be removed?

src/crypto/auth.sagas.test.ts
```diff
@@ -1,18 +1,111 @@
...
+  });
+
+  it('puts an error if the settings key is invalid', async () => {
+    await expectSaga(loginWorker, login('foo'))
+      .provide([[call.fn(init), undefined]])
+      .provide([[call.fn(checkSettingsKey), false]])
+      .call(init, 'foo')
+      .call(checkSettingsKey)
+      .put(loginFailed(translateRaw('LOGIN_ERROR')))
+      .silentRun(5000);
```
FrederikBolding, line 71: Why the need for the `5000` param?

FrederikBolding (changes requested):

src/app/screens/__tests__/Home.test.tsx
Mrtenz: The mock store doesn't have any persisted keys

src/crypto/auth.sagas.test.ts
Mrtenz: For some reason the saga would otherwise time out

FrederikBolding: Are we missing a mock?

Mrtenz: Every function that is called by the saga is mocked using `provide`, so I'm not sure why it times out

FrederikBolding (approved):

PR #97, opened 2021-05-03
Add basic loading screen + Fix some persistence issues

FrederikBolding:


FrederikBolding:
[ch8527]

src/api/store/settings.sagas.ts
```diff
@@ -27,6 +28,8 @@ export function* fetchSettingsWorker({ payload }: PayloadAction<string>) {
   const ciphertext: string = yield call(getFromStore, payload);
   if (ciphertext) {
     yield put(decryptSettings({ key: payload, value: ciphertext }));
+  } else {
+    yield put(rehydrateState({ key: payload, state: undefined }));
```
Mrtenz, line 32: I'm assuming this is for checking if the state was rehydrated? Maybe just add an action that adds the keys to `rehydratedKeys`, instead of resetting to initial state.

src/common/store/storage/persistence.slice.ts
```diff
@@ -37,7 +38,7 @@ export const getPersistenceState = createSelector(
 );
 
 export const getPersistentKeys = createSelector(
-  (state: Record<string, PersistState>) => state,
+  (state: Record<string, Optional<PersistState>>) => state,
```
Mrtenz, line 41: What's the reason for this? If necessary, can we use `Partial` (default in TS) instead of `Optional`?

Mrtenz (changes requested):

src/common/store/storage/persistence.slice.ts
FrederikBolding: Without this the typing of the state breaks for some selectors. Partial doesn't work for some reason. You can probably reproduce locally.

src/api/store/settings.sagas.ts
FrederikBolding: This is already the case for this action. Do we want a separate one that doesn't actually rehydrate?

Mrtenz (approved):

PR #98, opened 2021-05-03
Fix bug with having multiple permissions of the same origin

FrederikBolding:
Forgotten edgecase, needs testing

### DIFF
--- a/src/app/screens/RequestPermission.tsx
+++ b/src/app/screens/RequestPermission.tsx
@@ -7,8 +7,8 @@ import { ROUTE_PATHS } from '@app/routing';
 import {
   denyPermission,
   getPermissionRequest,
-  getPermissions,
   grantPermission,
+  hasExistingPermission,
   updatePermission
 } from '@common/store';
 import { translateRaw } from '@common/translate';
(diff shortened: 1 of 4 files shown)


src/app/screens/RequestPermission.tsx
```diff
@@ -4,16 +4,27 @@ import { useSelector } from 'react-redux';
...
 import { useDispatch } from '@store';
 
 export const RequestPermission = () => {
   const dispatch = useDispatch();
+  const permissions = useSelector(getPermissions);
   const request = useSelector(getPermissionRequest);
 
   const handleAllow = () => {
-    dispatch(grantPermission(request));
+    if (permissions.find((p) => p.origin === request.origin)) {
```
Mrtenz, line 23: Maybe move this logic to a selector?

src/app/screens/__tests__/RequestPermission.test.tsx
```diff
@@ -50,6 +50,21 @@ describe('RequestPermission', () => {
     );
   });
 
+  it('dispatches grantPermission on allow with existing perms', async () => {
```
Mrtenz, line 53:
```suggestion
  it('dispatches updatePermission on allow with existing perms', async () => {
```

Mrtenz (changes requested):

Mrtenz (approved):

PR #103, opened 2021-05-05
Sign Windows builds

GuillaumeRx:


forge.config.js
```diff
@@ -17,9 +17,15 @@ module.exports = {
   makers: [
     {
       name: '@electron-forge/maker-squirrel',
-      config: {
-        name: 'signer'
-      }
+      config: process.env.SHOULD_SIGN
```
FrederikBolding, line 20: Wouldn't this work?
`{ name: 'signer', ...(process.env.SHOULD_SIGN ? { certificateFile: bla } : {}) }`

FrederikBolding (changes requested):

forge.config.js
Mrtenz: Is the conditional needed at all? Wouldn't it just ignore it if `WINDOWS_PFX_FILE` is `undefined`?

FrederikBolding: 🤷‍♂️

GuillaumeRx: Yes you're right @Mrtenz !

Mrtenz (approved):

FrederikBolding (approved):

PR #104, opened 2021-05-05
Small polish changes

FrederikBolding:


src/app/components/TxQueue/TxQueueCard.tsx
```diff
@@ -22,12 +22,15 @@ export const TxQueueCard = ({ item }: { item: TxQueueEntry }) => {
   const symbol = chain?.nativeCurrency?.symbol ?? '?';
 
   return (
-    <Box pb="16px">
+    <Box pt={first ? '0' : '16px'} pb="16px">
```
Mrtenz, line 25: Is it possible to use `:first-of-type` or `:first-child` here, instead of passing stuff as React prop?

Mrtenz (changes requested):

src/app/components/TxQueue/TxQueueCard.tsx
FrederikBolding: If it is possible, I dont know how. But we need the prop anyway, so I don't see an issue with doing this?

Mrtenz (approved):

FrederikBolding:
[ch8560]

PR #110, opened 2021-05-10
Transaction Editing

FrederikBolding:


src/app/components/TxDetailsEdit.tsx
```diff
@@ -0,0 +1,88 @@
...
+const InputRow = ({
+  label,
+  unit,
+  ...props
+}: { label: string; unit?: string } & ComponentProps<typeof FormInput>) => (
+  <Row
+    label={label}
+    value={
+      <Box variant="horizontal-start" width="45%">
+        {/* @ts-ignore for now */}
```
Mrtenz, line 26: Why?

src/app/screens/EditTransaction.tsx
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+    try {
+      const tx = fromHumanReadable(form.values);
+      const queueItem = { ...currentTx, tx, adjustedNonce: false, userEdited: true };
+      dispatch(update(queueItem));
+      dispatch(selectTransaction(queueItem));
+      dispatch(replace(ROUTE_PATHS.TX));
+    } catch (err) {
+      console.error(err);
```
Mrtenz, line 64: This should be handled better.

src/app/screens/__tests__/EditTransaction.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+    },
+    transactions: {
+      queue: [transactionRequest],
+      currentTransaction: transactionRequest
+    }
+  });
+
+  const component = getComponent(mockStore);
+  return { component, mockStore };
+};
```
Mrtenz, line 43: Is there a reason why this can't be merged with `getComponent`?

Mrtenz (changes requested):

src/app/screens/EditTransaction.tsx
FrederikBolding: What would you prefer? It can be possibly be removed in favor of the form validation?

Mrtenz: I'd say that if form validation passes, the numbers should be valid BNs. Otherwise the form validation should be updated.

src/app/components/TxDetailsEdit.tsx
FrederikBolding: Still having type errors even when doing that 😢

Mrtenz: Yes, but those should be easier to fix.

FrederikBolding: Finally fixed it 😆

FrederikBolding:
[ch8565]

src/app/screens/EditTransaction.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const recipientAccount = tx && accounts.find((a) => a.address === tx.to);
+  const info = useSelector(getTransactionInfoBannerType);
+
+  const form = useForm(toHumanReadable(tx), yupValidator(SCHEMA), true);
+
+  const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    await form.validate();
+    if (form.error) {
+      return;
```
Mrtenz, line 55: Can you add a test for this?

src/app/screens/EditTransaction.tsx
```diff
@@ -0,0 +1,94 @@
...
+              recipient={tx.to && { address: tx.to, label: recipientAccount?.label }}
+            />
+            <Box variant="horizontal-start">
+              <Body fontSize="14px" color="BLUE_GREY" mb="2" mt="2">
+                {translateRaw('REQUEST_ORIGIN', { $origin: origin ?? translateRaw('UNKNOWN') })}{' '}
+                <TimeElapsed value={timestamp} />
+              </Body>
+              <LinkApp href="#" variant="barren" ml="auto">
+                <Image src={edit} height="20px" width="20px" />
+              </LinkApp>
```
Mrtenz, line 81: Why is this a link if it doesn't (seem to) do anything?

Mrtenz (changes requested):

src/app/components/ValidatedListener.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useAnyListener } from 'typed-react-form';
+
+export const ValidatedListener = <T,>({
+  form,
+  render
+}: {
+  form: FormState<T>;
+  render(values: T): ReactNode;
+}) => {
+  const [validatedValues, setValidatedValues] = useState(Object.assign({}, form.values));
```
Mrtenz, line 14:
```suggestion
  const [validatedValues, setValidatedValues] = useState({ ...form.values });
```

Mrtenz (changes requested):

Mrtenz (approved):

PR #111, opened 2021-05-10
Implement new paper wallet design

Mrtenz:


src/app/components/PaperWallet.tsx
```diff
@@ -1,102 +1,216 @@
...
+          type === WalletType.PRIVATE_KEY
+            ? translateRaw('PRIVATE_KEY')
+            : translateRaw('MNEMONIC_PHRASE')
+        }
+      >
+        <Box>
+          <QRCode size={125} level="M" value={secret} />
+        </Box>
+        <Body fontSize="10px" lineHeight="12px">
+          {secret}
```
FrederikBolding, line 133: With private keys this overflows outside the box

src/app/components/PaperWallet.tsx
```diff
@@ -1,102 +1,216 @@
...
+        </Flex>
+      </Block>
+    </GridBox>
+    <GridBox>
+      <Flex variant="vertical-center" justifyContent="space-between" pt="45px" height="100%">
+        <Box sx={{ textAlign: 'center' }}>
+          <Body fontWeight="bold" fontSize="20px" lineHeight="24px" mb="3">
+            {translateRaw('KEEPING_YOUR_CRYPTO_SAFE')}
+          </Body>
+          <Image src={safeWallet} alt="Safe wallet" />
```
FrederikBolding, line 186: Some of the images are missing on the .png

src/app/screens/AddAccount/AddAccountBackup.tsx
```diff
@@ -1,35 +1,53 @@
...
 
   const handleToggle = () => setPersistent((value) => !value);
 
   const handleAdd = () => {
     dispatch(addSavedAccounts(persistent));
   };
 
+  useEffect(() => {
+    if (paperWallet.current) {
+      toPng(paperWallet.current).then(setPaperWalletImage);
```
FrederikBolding, line 31: Seems to freeze sometimes

src/common/store/accounts.slice.ts
```diff
@@ -119,7 +136,7 @@ export const getAccounts = createSelector(
 
 export const getAccountsToAdd = createSelector(
   (state: { accounts: AccountsState }) => state.accounts,
-  (accounts) => accounts.accountsToAdd
+  (accounts) => accounts.add
```
FrederikBolding, line 139: Why the rename?

src/app/components/PaperWallet.tsx
```diff
@@ -1,102 +1,216 @@
...
+      borderLeft: !last && inverted && '2px dashed',
+      borderColor: 'border.light',
+      transform: inverted && 'rotate(180deg)'
+    }}
+  >
+    {children}
+  </Box>
+);
+
+export const PaperWallet = forwardRef(({ type, address, secret }: PaperWalletProps, ref) => (
```
FrederikBolding, line 109: Should we add the DPath for mnemonic paper wallets?

src/common/store/accounts.slice.ts
```diff
@@ -8,15 +8,22 @@ import type {
...
 } from '@types';
 
 export interface AccountsState {
   accounts: IAccount[];
-  accountsToAdd: SerializedWalletWithAddress[];
   addresses: GetAddressesResult[];
   isFetching: boolean;
   fetchError?: string;
+  add?: {
+    // Keystore files can be added as well, but for saving paper wallets we only support mnemonics
```
FrederikBolding, line 21: Is there a reason to distinguish here? The secret of a keystore is the private key? Couldn't it still be typed as a `KEYSTORE`?

src/crypto/accounts.slice.ts
```diff
@@ -81,11 +89,29 @@ export function* fetchAccount(wallet: SerializedWallet): SagaIterator<Serialized
 
 export function* fetchAccountsWorker({ payload }: PayloadAction<SerializedWallet[]>) {
   try {
-    const wallets: SerializedWalletWithAddress[] = yield all(
+    const type =
```
FrederikBolding, line 92: These two lines seem to be complicating things too much, can we abstract away?

FrederikBolding (changes requested):

src/app/components/PaperWallet.tsx
Mrtenz: Maybe? That's a good question for design probably.

FrederikBolding: Agreed. I asked them about this.

src/crypto/accounts.slice.ts
```diff
@@ -44,21 +42,31 @@ const slice = createSlice({
...
 });
 
-export const { setAccountsToAdd } = slice.actions;
+export const { setAddAccounts, clearAddAccounts } = slice.actions;
 
 export const addSavedAccounts = createAction<boolean>(`${sliceName}/addSavedAccounts`);
 
 export default slice;
 
 export const getAccountsToAdd = createSelector(
```
FrederikBolding, line 67: Let's not export this to prevent confusion with the other function of the same name.

src/crypto/accounts.slice.ts
FrederikBolding: Maybe moving it to two separate functions? We can use a switch statement instead for the secret.

In general I just want it to be clearer that it is only used for paper wallets. Perhaps with some more comments?

src/common/store/accounts.slice.ts
FrederikBolding: Can we come up with a better name? 🤔

src/common/store/accounts.slice.ts
FrederikBolding: Wouldn't it be clearer to keep the details of that in the PaperWallet implementation? I.e. supporting all types but inside PaperWallet `KEYSTORE` = `PRIVATE_KEY`?

src/app/components/QR.tsx
```diff
@@ -0,0 +1,27 @@
...
+        margin: 0
+      })
+    );
+  }, [data]);
+
+  if (image) {
+    return <Image src={image} alt="QR" size={size} />;
+  }
+
+  return null;
```
FrederikBolding, line 26: Wouldn't this render `null`?

FrederikBolding (changes requested):

src/common/store/accounts.slice.ts
Mrtenz: Like what?

FrederikBolding (approved):

PR #112, opened 2021-05-12
Add extended key banner to mnemonic flow

Mrtenz:


Mrtenz:
[ch8441]

src/app/components/Banner.tsx
```diff
@@ -0,0 +1,92 @@
...
+const icons: { [key in BannerType]: string } = {
+  success,
+  info,
+  action,
+  warning,
+  error
+};
+
+export const InnerBanner = ({ type, children }: PropsWithChildren<Pick<BannerProps, 'type'>>) => (
+  <Box backgroundColor={theme.variants.banner[type].color} sx={{ borderRadius: 'banner' }}>
```
FrederikBolding, line 31: This border radius is not defined.

src/app/components/Banner.tsx
```diff
@@ -0,0 +1,92 @@
...
+            sx={{ textTransform: 'uppercase' }}
+          >
+            {label}
+          </Body>
+        </Flex>
+        <Flex variant="horizontal-start">
+          {banner && <InnerBanner type={type}>{banner}</InnerBanner>}
+          {children && (
+            <Box ml="2" sx={{ userSelect: 'none' }}>
+              <Image src={caret} alt="Caret" sx={{ transform: isExtended && 'rotate(180deg)' }} />
```
FrederikBolding, line 80: This image is not centered.

src/app/components/Banner.tsx
FrederikBolding, line 31: We should use the theme from the provider here. Otherwise it will make it difficult to change down the road

FrederikBolding (changes requested):

FrederikBolding (approved):

PR #114, opened 2021-05-17
Auto focus when user needs to interact with the app

FrederikBolding:


FrederikBolding:
[ch8583]

src/crypto/accounts.slice.test.ts
```diff
@@ -17,6 +16,7 @@ import {
   setGeneratedAccount
 } from '@common/store';
 import { DEFAULT_MNEMONIC_INDEX } from '@config';
+import { createWallet, derivePrivateKey, getAddress, getAddresses } from '@crypto/crypto';
```
Mrtenz, line 19: You can use a relative import here (and in some other files as well).

src/window.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const showWindowOnTop = (window: BrowserWindow) => {
+  window.setAlwaysOnTop(true);
+  // Hack to show on current desktop - see: https://github.com/electron/electron/issues/5362
+  window.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
+  window.show();
+  window.focus();
+  window.setVisibleOnAllWorkspaces(false);
+  window.setAlwaysOnTop(false);
+};
```
Mrtenz, line 11: What's the reason for this being a separate file?

Mrtenz (changes requested):

src/window.ts
FrederikBolding: Just for reusability. It is used in `main.ts` and the saga.

Mrtenz: Can you move it to `utils` then?

FrederikBolding: Sure

Mrtenz (approved):

PR #121, opened 2021-05-24
Refactor IAccount type to only include path and index for mnemonics

Mrtenz:


Mrtenz:
[ch8561]

src/app/screens/SignTransaction/SignTransaction.tsx
```diff
@@ -36,9 +36,22 @@ export const SignTransaction = () => {
   }
 
   if (currentAccount) {
+    if (currentAccount.type === WalletType.MNEMONIC) {
```
FrederikBolding, line 39: Why is this necessary? This was much cleaner without this.

FrederikBolding (changes requested):

src/app/screens/SignTransaction/SignTransaction.tsx
Mrtenz: To properly infer the type of `currentAccount`. Otherwise, the type would be inferred as `never`. Alternatively I can do a type cast if you think that's better?

FrederikBolding: I would prefer that over this yeah.

FrederikBolding (approved):

PR #133, opened 2021-06-29
Initial signer/extension communication through WebSockets + JSON-RPC

Mrtenz:


packages/extension/src/relay/handlers/content.ts
```diff
@@ -0,0 +1,25 @@
...
+      return;
+    }
+
+    const response = await sendMessage(event.data);
+    if (
+      !is(response, RelayResponseStruct) ||
+      response.id !== event.data.id ||
+      response.target !== RelayTarget.Content
+    ) {
+      return;
```
FrederikBolding, line 20: Can we test the error state?

packages/extension/src/store/jsonrpc.slice.ts
```diff
@@ -0,0 +1,31 @@
...
+const initialState: JsonRpcState = {
+  network: {
+    providers: ['https://api.mycryptoapi.com/eth'],
+    chainId: 1
+  }
+};
+
+const sliceName = 'jsonrpc';
+
+const slice: Slice<JsonRpcState> = createSlice({
```
FrederikBolding, line 20: This slice looks to not be in use?

packages/extension/src/store/sagas.ts
```diff
@@ -0,0 +1,7 @@
+import { all } from 'redux-saga/effects';
+
+import { socketsSaga } from './sockets.slice';
+
+export default function* rootSaga(/* ipc: ReduxIPC */): Generator {
```
FrederikBolding, line 5: Can we remove the parameter comment here?

packages/extension/src/store/sockets.slice.ts
```diff
@@ -0,0 +1,192 @@
...
+    takeEvery(handleRequest.type, handleRequestWorker),
+    takeLatest(setConnected.type, setConnectedWorker)
+  ]);
+}
+
+/**
+ * Creates a new random private key to be used for signing requests.
+ */
+export function* createPrivateKeyWorker() {
+  const privateKey: string = yield call(createRandomPrivateKey);
```
FrederikBolding, line 76: Can we test this saga?

packages/extension/src/types/relay.ts
```diff
@@ -0,0 +1,47 @@
+import type { Infer } from 'superstruct';
+import { any, array, number, object, optional, pattern, string, union, unknown } from 'superstruct';
+
+export enum RelayTarget {
+  Page = 'Page',
+  Content = 'Content',
+  Background = 'Background',
+  External = 'External'
```
FrederikBolding, line 8: What's this needed for?

packages/extension/src/relay/background.test.ts
```diff
@@ -0,0 +1,70 @@
+import { generateUUID } from '@signer/common';
+import { chrome } from 'jest-chrome';
+
+import { RelayTarget } from '../types';
+import { createBackgroundMessageSender } from './background';
+
+const sleep = () => new Promise((resolve) => setTimeout(resolve, 1));
```
FrederikBolding, line 7: Can we share this code between the files that use it?

packages/extension/src/store/sockets.slice.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      .fork(socketRequestWorker, instance);
+
+    callback({ data: JSON.stringify({ foo: 'bar' }) } as MessageEvent);
+
+    saga.put(message({ foo: 'bar' }));
+
+    await promise;
+  });
+
+  it('reconnects if the connection is closed', async () => {
```
FrederikBolding, line 134: How does this test work? 😆

packages/extension/src/store/sockets.slice.ts
```diff
@@ -0,0 +1,192 @@
...
+  try {
+    while (true) {
+      try {
+        const action: MessageEvent = yield take(channel);
+        const json = JSON.parse(action.data);
+
+        yield put(message(json));
+      } catch (error) {
+        yield put(setConnected(false));
+        channel.close();
```
FrederikBolding, line 167: Any idea why these aren't these covered?

FrederikBolding (changes requested):

packages/extension/src/relay/handlers/content.ts
Mrtenz: What do you mean? This is not an error state

packages/extension/src/utils/inject.ts
```diff
@@ -0,0 +1,12 @@
+export const injectScript = (): Promise<void> => {
+  return new Promise((resolve) => {
+    const script = document.createElement('script');
+    script.src = chrome.runtime.getURL('lib/page.js');
+    script.addEventListener('load', () => {
+      script.remove();
+      resolve();
```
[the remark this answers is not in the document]
Mrtenz: I have no idea how we would test this.

packages/extension/src/store/jsonrpc.slice.ts
Mrtenz: It's not in use right now, but will be needed for handling requests to an actual JSON-RPC node.

packages/extension/src/types/relay.ts
Mrtenz: Two things:
- Depending on how messages are sent across browser extensions, they may not be fully isolated from other extensions. This basically checks if messages sent are coming from a part of the extension.
- This ensures that messages aren't replayed twice. Calling `window.postMessage` in the content script will also trigger the event listener in the content script.

packages/extension/src/store/sockets.slice.test.ts
Mrtenz: It gets the `close` event callback (`addEventListener.mock.calls[2][1]`) and calls it manually.

packages/extension/src/utils/inject.ts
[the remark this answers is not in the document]
FrederikBolding: 👍

packages/extension/src/relay/handlers/content.ts
FrederikBolding: Sorry, can we test this edge case? I just mean that this line isn't covered 😄

packages/extension/src/types/relay.ts
FrederikBolding: Sorry, I specifically meant `External`. I couldn't find it used anywhere?

FrederikBolding (approved):

PR #140, opened 2021-09-03
Fix potential security issues in origin validation

FrederikBolding:
[ch8972]

### DIFF
--- a/packages/signer/src/api/ws.sagas.ts
+++ b/packages/signer/src/api/ws.sagas.ts
@@ -156,7 +156,7 @@ export function* handleRequest({ socket, request, data }: WebSocketMessage) {
 
   const origin = getOriginHost(request);
   // Dont serve requests without origin
-  if (origin === undefined || origin === null) {
+  if (origin == null) {
     // @todo Decide what error message to use
     return socket.send(
       JSON.stringify(
--- a/packages/signer/src/api/ws.sagas.ts
+++ b/packages/signer/src/api/ws.sagas.ts
@@ -138,6 +138,14 @@ export function* waitForPermissions(permission: Permission) {
   }
 }
 
+const getOriginHost = (request: IncomingMessage) => {
+  try {
+    return request.headers.origin && new URL(request.headers.origin).host;
+  } catch {
+    return null;
+  }
+};
+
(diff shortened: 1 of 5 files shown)


packages/signer/src/api/ws.sagas.ts
```diff
@@ -146,15 +154,41 @@ export function* handleRequest({ socket, request, data }: WebSocketMessage) {
 
   const { signature, publicKey, ...jsonRpcRequest } = fullRequest;
 
-  const permissions: Permission[] = yield select(getPermissions);
-  const origin = request.headers.origin && new URL(request.headers.origin).host;
+  const origin = getOriginHost(request);
+  // Dont serve requests without origin
+  if (origin === undefined || origin === null) {
```
Mrtenz, line 159:
```suggestion
  if (origin == null) {
```
or just
```suggestion
  if (!origin) {
```

packages/signer/src/utils/hashing.ts
```diff
@@ -17,6 +17,10 @@ export const signRequest = async (data: JsonRPCRequest, privateKey: string) => {
 };
 
 export const verifyRequest = async (sig: string, data: JsonRPCRequest, publicKey: string) => {
-  const hash = await hashRequest(data);
-  return verify(sig, hash, publicKey);
+  try {
+    const hash = await hashRequest(data);
+    return await verify(sig, hash, publicKey);
```
Mrtenz, line 22: `await` is redundant when using `return`.

Mrtenz (changes requested):

packages/signer/src/utils/hashing.ts
FrederikBolding: Not when trying to catch the promise rejection!

packages/signer/src/api/ws.sagas.ts
FrederikBolding: I would prefer being explicit in this case if it is fine with you.

Mrtenz: Then using the first option should be fine, right?

Mrtenz (approved):

PR #143, opened 2021-09-14
Enable renderer sandbox

Mrtenz:
This enables the sandbox for all renderer processes. Since we don't use Node.js within the renderer process, this doesn't seem to impact any functionality.


Mrtenz:
[ch8975]

FrederikBolding (approved):

cure53 (approved):

PR #147, opened 2021-09-24
Use scrypt with random salt for key derivation

Mrtenz:


Mrtenz:
[sc-8978]

packages/signer/src/utils/encryption.spec.ts
```diff
@@ -31,7 +32,7 @@ describe('createEncryptionKey', () => {
 
 describe('hashPassword', () => {
   it('correctly SHA256 hashes a string', async () => {
```
FrederikBolding, line 34: Update this to be correct

packages/signer/src/utils/encryption.ts
```diff
@@ -52,10 +46,20 @@ export const decrypt = (data: string, key: Buffer) => {
...
-  return pbkdf2(password, SALT, ITERATIONS, KEY_LENGTH, HASH_ALGORITHM);
+  // `util.promisify` doesn't properly infer the types of Scrypt, so it's wrapped in a promise here
+  // instead
+  return new Promise((resolve, reject) => {
+    crypto.scrypt(password, salt, KEY_LENGTH, (error, key) => {
+      if (error) {
+        return reject(error);
+      }
+
+      resolve(key);
```
FrederikBolding, line 62: Missing return?

packages/signer/src/crypto/secrets.test.ts
```diff
@@ -127,10 +136,14 @@ describe('hasSettingsKey', () => {
 
 describe('checkSettingsKey', () => {
   it('checks if the settings key can be decrypted', async () => {
-    (keytar.getPassword as jest.MockedFunction<typeof keytar.getPassword>).mockImplementation(
-      async () =>
-        '6a7cd51df01e5152968a5acbd312c2d55c08eb7cf4061c1a4d1de326ebcbbea026d35cd7b5c43ade2d21938ada7a165c39c8f3fd'
-    );
+    (keytar.getPassword as jest.MockedFunction<typeof keytar.getPassword>)
+      .mockImplementationOnce(
+        async () => 'dc509bcfc343f1bebb4d75749695fd3ef204306f07eff6f86eec91587ba03bbc'
```
FrederikBolding, line 141: Can you define this globally if it is re-used for all tests?

packages/signer/src/crypto/secrets.ts
```diff
@@ -19,7 +19,7 @@ export const clearEncryptionKey = () => {
 };
 
 export const init = async (password: string) => {
-  const key = await hashPassword(password);
+  const key = await hashPassword(password, await getSalt());
```
FrederikBolding, line 22: Maybe add a const for the salt for readability?

FrederikBolding (changes requested):

FrederikBolding (approved):
lgtm - ask cure53 if we should be tweaking scrypt params

PR #148, opened 2021-09-24
Improved nonce handling

FrederikBolding:


FrederikBolding:
[sc-8973]

packages/common/src/store/transactions.slice.ts
```diff
@@ -94,15 +93,17 @@ export const getNoncesInQueue = createSelector(
...
-  );
+export const hasNonceConflict = createSelector(
+  getApprovedTransactions,
+  getCurrentTransaction,
+  (transactions, currentTx) =>
+    transactions.find(
+      (h) =>
+        h.tx.from === currentTx.tx.from &&
+        h.tx.chainId === currentTx.tx.chainId &&
+        bigify(h.tx.nonce).eq(bigify(currentTx.tx.nonce))
```
Mrtenz, line 104: Why is bigify necessary here? Can the nonce be in a different format?

packages/common/src/translations/lang/en.json
```diff
@@ -165,7 +165,8 @@
     "REVIEW_SECURITY_DETAILS": "Review Security Details",
     "NONCE_CHANGED": "Transaction nonce has been changed",
     "NONCE_CHANGED_CONTENT": "This transaction had the same nonce as a previously sent transaction. We've changed it for you, but you may edit it if you need to, using the pencil icon below.",
-    "NONCE_CONFLICT_IN_QUEUE": "Nonce is used in another transaction",
+    "NONCE_CONFLICT_IN_QUEUE": "The selected nonce is used by another TX in your queue. [Read More](https://support.mycrypto.com/general-knowledge/ethereum-blockchain/what-is-nonce?utm_medium=organic&utm_source=quill&utm_campaign=txrequest)",
```
Mrtenz, line 168: Can we use "transaction" instead of "TX"?

packages/extension/src/utils/jsonrpc.ts
```diff
@@ -11,28 +13,37 @@ export const toJsonRpcRequest = (message: RelayMessage): JsonRPCRequest => ({
...
   if (!Array.isArray(params)) {
     return undefined;
   }
 
-  return [{ ...((params[0] as Record<string, unknown>) ?? {}), chainId }];
+  const providers = state.jsonrpc.network.providers;
+  const provider = new FallbackProvider(providers.map((url) => new JsonRpcProvider(url)));
+
+  // @ts-expect-error No type atm
+  const nonce = params[0].nonce ?? hexlify(await provider.getTransactionCount(params[0].from));
```
Mrtenz, line 28: Can you explain?

Mrtenz (changes requested):

packages/extension/src/utils/jsonrpc.ts
FrederikBolding: `params` is typed as `unknown` for this function currently

packages/common/src/store/transactions.slice.ts
FrederikBolding: Just to be safe as it is a Redux value it may be serialized, right?

packages/extension/src/utils/jsonrpc.ts
Mrtenz: Can you use a type guard?

packages/extension/src/utils/jsonrpc.ts
```diff
@@ -24,8 +32,10 @@ export const addMissingParams = async (
   const providers = state.jsonrpc.network.providers;
   const provider = new FallbackProvider(providers.map((url) => new JsonRpcProvider(url)));
 
-  // @ts-expect-error No type atm
-  const nonce = params[0].nonce ?? hexlify(await provider.getTransactionCount(params[0].from));
+  const nonce =
+    params[0].nonce ?? params[0].from
+      ? hexlify(await provider.getTransactionCount(params[0].from!))
+      : null;
```
Mrtenz, line 38: Can you move this into a separate function to make it more readable?

Mrtenz (changes requested):

Mrtenz (approved):

PR #150, opened 2021-09-27
Auto redirect for incoming tx

FrederikBolding:


FrederikBolding:
[sc-9033]

Mrtenz (approved):
Tested on Linux, works fine

PR #155, opened 2021-10-01
Add replay protection to JSON-RPC

Mrtenz:
A nonce is stored in the application for each public key, and must be incremented by one for each JSON-RPC request.

### DIFF
--- a/packages/signer/src/api/ws.persist.ts
+++ b/packages/signer/src/api/ws.persist.ts
@@ -1,0 +1,10 @@
+import type { PersistConfig } from '@signer/common';
+import { createPersistReducer, incrementNonce, wsSlice } from '@signer/common';
+
+const persistConfig: PersistConfig = {
+  key: wsSlice.name,
+  whitelistedActions: [incrementNonce.type],
+  whitelistedKeys: []
+};
+
+export const persistedReducer = createPersistReducer(persistConfig, wsSlice.reducer);
--- a/packages/signer/src/api/ws.persist.ts
+++ b/packages/signer/src/api/ws.persist.ts
@@ -4,7 +4,7 @@ import { createPersistReducer, incrementNonce, wsSlice } from '@signer/common';
 const persistConfig: PersistConfig = {
   key: wsSlice.name,
   whitelistedActions: [incrementNonce.type],
-  whitelistedKeys: []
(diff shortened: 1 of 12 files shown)


packages/signer/src/api/ws.sagas.ts
```diff
@@ -162,22 +184,34 @@ export function* handleRequest({ socket, request, data }: WebSocketMessage) {
...
+          error: { code: '-32600', message: 'Invalid request' }
+        })
+      )
+    );
+  }
+
+  // Dont serve requests with invalid nonces
+  const isValidNonce: boolean = yield call(verifyRequestNonce, jsonRpcRequest, publicKey);
+  if (!isValidNonce) {
+    return socket.send(
```
FrederikBolding, line 210: Can we test this?

packages/signer/src/api/ws.sagas.ts
```diff
@@ -146,6 +149,25 @@ const getOriginHost = (request: IncomingMessage) => {
   }
 };
 
+/**
+ * Verifies the JSON-RPC request nonce based on the ID and public key. If the nonce is valid, the
```
FrederikBolding, line 153: Perhaps add something about: "we expect the JSON RPC ids to be nonces and increment for each request..."?

FrederikBolding (changes requested):

packages/signer/src/api/ws.persist.ts
```diff
@@ -0,0 +1,10 @@
+import type { PersistConfig } from '@signer/common';
+import { createPersistReducer, incrementNonce, wsSlice } from '@signer/common';
+
+const persistConfig: PersistConfig = {
+  key: wsSlice.name,
+  whitelistedActions: [incrementNonce.type],
+  whitelistedKeys: []
```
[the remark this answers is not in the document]
Mrtenz: Yes, I initially used
```ts
{
  [publicKey: string]: number;
}
```
as type for the slice, but then realised I couldn't whitelist the keys. Changed it to have a `nonces` key but forgot to add it to the array here...

FrederikBolding (approved):
lgtm

Mrtenz:
[sc-8974]

PR #158, opened 2021-10-01
Add accounts empty state

FrederikBolding:


FrederikBolding:
[sc-8605]

packages/signer/src/app/screens/Accounts.tsx
```diff
@@ -82,7 +85,26 @@ export const Accounts = () => {
...
+          <Body variant="muted">{translateRaw('ACCOUNTS_EMPTY_BODY')}</Body>
+        </Box>
+      </Container>
+      <PanelBottom variant="clear">
+        <LinkApp href={ROUTE_PATHS.ADD_ACCOUNT}>
+          <Button>{translateRaw('MENU_ADD_ACCOUNT')}</Button>
+        </LinkApp>
+      </PanelBottom>
+    </>
+  ) : (
```
Mrtenz, line 107: Maybe do something like this to make it slightly more readable?
```ts
if (accounts.length === 0) {
  return ...;
}

return ...;
```

Mrtenz (changes requested):

packages/signer/src/app/screens/Accounts.tsx
FrederikBolding: Sure

Mrtenz (approved):

PR #179, opened 2021-12-02
Fix Snyk warnings

FrederikBolding:
- Bump Electron
- Remove Spectron as it is being deprecated


FrederikBolding:
[sc-9135]

packages/quill/src/main.ts
```diff
@@ -54,18 +54,15 @@ const createWindow = (): void => {
...
+    frame: false,
     transparent: true,
     webPreferences: {
       devTools: isDev,
       preload: MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY,
       // For security reasons the following params should not be modified
       // https://electronjs.org/docs/tutorial/security#isolation-for-untrusted-content
       nodeIntegration: false,
-      contextIsolation: true,
-      worldSafeExecuteJavaScript: true
```
Mrtenz, line 68: Is this safe to remove?

FrederikBolding: Yes. It has been removed and the behaviour is enabled by default: https://www.electronjs.org/blog/electron-14-0#removed-worldsafeexecutejavascript

Mrtenz (approved):

PR #186, opened 2022-01-11
Bump Ramda

FrederikBolding:


FrederikBolding:
This should fix the vuln, but seems that Snyk isn't updated yet: https://github.com/ramda/ramda/releases/tag/v0.27.2

Mrtenz (approved):

PR #193, opened 2022-01-28
Consolidate no accounts added view

FrederikBolding:


FrederikBolding:
[sc-9484]

packages/quill/src/app/screens/Accounts.tsx
```diff
@@ -76,35 +74,16 @@ const Account = ({ account }: { account: IAccount }) => {
...
-          </LinkApp>
-        </PanelBottom>
-      </>
-    );
-  }
+  useEffect(() => {
+    if (accounts.length === 0) {
+      dispatch(push(ROUTE_PATHS.SETUP_ACCOUNT));
+    }
+  });
```
Mrtenz, line 86: Should this be dependant on `accounts.length`?

Mrtenz (changes requested):

packages/quill/src/app/screens/Accounts.tsx
FrederikBolding: Yep. Good call

Mrtenz (approved):

PR #202, opened 2022-02-10
View Signed TX

FrederikBolding:


FrederikBolding:
[sc-9513]

packages/quill/src/app/screens/Transaction.tsx
```diff
@@ -97,11 +103,17 @@ export const Transaction = () => {
...
+      {result === TxResult.APPROVED && isHistoryTx(currentTx) && (
         <PanelBottom py="24px">
-          <Body fontWeight="bold" textAlign="center">
-            {translateRaw('TRANSACTION_APPROVED_TIP')}
-          </Body>
+          {offline || getMinuteDifference(Date.now(), currentTx.actionTakenTimestamp) >= 5 ? (
+            <Button variant="inverted" onClick={handleViewSigned}>
+              {translateRaw('VIEW_SIGNED_TX')}
+            </Button>
+          ) : (
```
Mrtenz, line 112: Do we only want to show the button after 5 minutes? My understanding was that we'd show both the button and the message, but hide the message after 5 minutes.

Mrtenz (changes requested):

Mrtenz (approved):

PR #205, opened 2022-02-15
Show error message on invalid input

Mrtenz:
[sc-9478]


packages/quill/src/app/components/TxDetailsEdit.tsx
```diff
@@ -36,45 +36,55 @@ type InputRowProps = {
...
+                }
+              `}
+            />
+            {unit && (
+              <Body ml="-3rem" sx={{ zIndex: 2 }}>
+                {unit}
+              </Body>
+            )}
+          </Box>
+          {error && <Body variant="error">{translateRaw('INVALID_INPUT')}</Body>}
```
FrederikBolding, line 75: Can we add this for the `data` field too?

FrederikBolding (changes requested):

FrederikBolding (approved):

